PR #299, opened 2014-12-02
Customise weight/lengths measurement

AntonyThorpe:
The problem we face is that our inventory system measures lengths in metres and SS Shop currently uses centimetres.  This pull request proposes an additional decimal place for the length measurement to enable accuracy down to millimetres, plus the ability to customise the CMS presentation through mysite/config.yml.

### DIFF
--- a/code/product/Product.php
+++ b/code/product/Product.php
@@ -22,10 +22,10 @@ class Product extends Page implements Buyable {
 		'BasePrice' => 'Currency', // Base retail price the item is marked at.
 
 		//physical properties
-		'Weight' => 'Decimal(9,2)',
-		'Height' => 'Decimal(9,2)',
-		'Width' => 'Decimal(9,2)',
-		'Depth' => 'Decimal(9,2)',
+		'Weight' => 'Float',
+		'Height' => 'Float',
+		'Width' => 'Float',
+		'Depth' => 'Float',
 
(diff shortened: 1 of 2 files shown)


jedateach:
Thanks for this @AntonyThorpe.

Increasing the granularity of the Decimal db fields is definitely needed if we'll be dealing in various different units. I wonder if we should go straight for Floating point numbers instead to allow for greatest flexibility.

Regarding the weight labels, these can actually be changed using a translation, eg:
```yaml
en:
   Product:
     HEIGHT: Height (cm)
```
But, I see that you've essentially finished the work, based on my TODOs - So fair enough thankyou :)
Could you squash the commits into a single commit?, and then I can merge.

---

For a more long-term/complete solution I imagine some kind of unit system / weight form and db fields could help with capturing and displaying data. You could capture in various kinds of units, and have them all stored in some base unit like "meters". Perhaps handling conversions between types of unit systems also.
I started some work once upon a time here: https://gist.github.com/jedateach/2503707

AntonyThorpe:
Commits squashed as requested.  Cheers.

jedateach:
:+1: 

@markguinn I assume the switch to "Float" will allow for greater flexiblity of weight types. Eg things like 2342kg 0.0001 kg or 123223.23523234 kg

PR #300, opened 2014-12-29
Fixes a checkout bug related to Member addresses

markguinn:
Given a single-page checkout and given the membership component comes after the address book components, the addresses don't get assigned to the member correctly. Supercedes #276.


jedateach:
:+1:

PR #301, opened 2014-12-29
Added hooks for ajax integration

markguinn:
This PR doesn't add full ajax support but it adds extension hooks to make that happen. In the mean time, so that these features can be used safely in modules, I've separated the actual features into a couple modules. I felt the ajax module itself could be useful outside of the shop module so I think we should probably leave it separate. If you want to merge the shop-ajax module into shop core to keep from having two sets of pull requests in the future, I'm totally fine with that.

- https://github.com/markguinn/silverstripe-ajax
- https://github.com/markguinn/silverstripe-shop-ajax

This is a resubmission of #275 because that PR failed Travis/Scrutinizer (due to timeouts). Since that issue has been resolved I'm resubmitting this.

### DIFF
--- a/code/cart/ShoppingCart.php
+++ b/code/cart/ShoppingCart.php
@@ -417,6 +417,12 @@ class ShoppingCart_Controller extends Controller{
 		return "";
 	}
 
+
+	/**
+	 * This is used here and in VariationForm and AddProductForm
+	 * @param bool|string $status
+	 * @return bool
+	 */
 	public static function direct($status = true) {
 		if(Director::is_ajax()){
 			return $status;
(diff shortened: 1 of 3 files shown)


jedateach:
Thanks for this Mark. Its looking to me like a good solution to the ajax abstraction discussion we had previously. 

Could you add the shop-ajax module to the composer 'suggests' block?

Regarding testing, do you think we should also write some behat tests to test ajax functionality? Do you have any experience there?
I guess we could also write some dummy extensions to return some dummy data to be asserted against in unit tests.

markguinn:
I've written a few behat tests but I'm no expert. I've got about 3 projects that all need to be finished this month so I doubt I'll have time soon to devote to this but I'm happy to pick it back up in a little while if that's ok with you.

I've just added shop-ajax. Do you want me to squash that second commit into a new PR? I'm not sure how to do it on the same PR.

jedateach:
Thats ok regarding tests. I haven't written a behat test myself yet.

You just force-push to the same branch, and it gets overridden.

jedateach:
:+1:

PR #306, opened 2015-02-05
Stripped out OrderStatusLog functionality.

jedateach:
This functionality has not been used, tested or maintained for some time, and can be added in at a later stage, perhaps by a sub-module.


wilr:
-1 We use sendStatusChange() for our store. If it is going to be stripped out then a module would be good to setup before it lands.

markguinn:
I've never used this feature, but it seems like it makes sense to have. How much work would it take to make this a first-class citizen? Is it just a matter of writings some tests? Is there currently a way to make the status notification changes happen automatically or do you have to write that yourself?

jedateach:
@wilr ok cool, thanks for sharing. Yeah, I agree that if this is part of 1.0 functionality, then we need to keep it in.

@markguinn I'm not sure. I've never used it either, but I have imagined that it would be automated.

A basic implementation could be that whenever the `Status` field changes, a new log entry is created. Its almost like the `Order` table should be versioned.
---
I think I'll continue pull requesting instead of just making changes to shop from now on. I already see its going to prevent hassles.

wilr:
@jedateach Yeah, happy to review PRs. Of course, my experience is limited to a couple of projects but will try and help anyway possible.

markguinn:
@wilr if you're actively using this could you fill us in on how you use it? Are you extending Order and calling sendStatusChange from onAfterWrite or something?

I agree that automatically created log entries seems like a good plan. I wonder too about having a config option on order that would be an array of statuses for which the status email gets sent automatically. If this was empty by default you'd maintain backwards compatibility but make using this feature simple to opt in on.

wilr:
@markguinn calling sendStatusChange() in a function which is called when an order has been shipped to send a notification to the user.

adrian-stein:
Hey guys, I assume making this code into a 'module' has still not been done. Just so it doesnt get removed before an equivalent module is made I also plan to use it. How, I am not sure but once I figure it out I will let you know.

adrian-stein:
ON a side note I found this module: [https://github.com/milkyway-multimedia/ss-shop-order-history](https://github.com/milkyway-multimedia/ss-shop-order-history) which looks like they hook into the class as well.

markguinn:
Oh, that milkyway-multimedia module looks brilliant. That might just solve this issue. We could deprecate everything here and remove it in 2.0 (assuming we evaluate that module in depth and determine that it does everything needed). Thanks for pointing that out @adrian-stein!

jedateach:
I'm not likely to work on this PR any time soon, so closing.

PR #307, opened 2015-02-05
Email refactor + improvement

jedateach:
I've added in a 'confirmation email' , which gets sent when an order is placed, but not yet paid for. Receipt is sent after payment.

Also added a AdminNotification email. The email notification to the admin sometimes needs to be different.

Refactored OrderProcessor email stuff into its own class.

Updated docs

### DIFF
--- a/code/checkout/OrderEmail.php
+++ b/code/checkout/OrderEmail.php
@@ -3,6 +3,7 @@
 /**
  * This class handles the receipt email which gets sent once an order is made.
  * You can call it by issuing sendReceipt() in the Order class.
+ * @deprecated 2.0 Use Email instead, and set template
  */
 class Order_ReceiptEmail extends Email {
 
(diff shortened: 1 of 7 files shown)


markguinn:
I like this. Breaking this out into it's own class is a good design decision. My only question would be - is it worth keeping the sendEmail methods in OrderProcessor to maintain backwards compatibility? They could be deprecated and just call out to OrderEmailNotifier to keep things DRY. Either way is fine by me, but if this merges as is I know I'll have to change some code in my current project so I'd expect others might as well.

jedateach:
Yup - I like the deprecation idea. I guess if we're going semantic versioning, then it'll be removed in version 2 of the module...does that sound right to you?

markguinn:
Sounds like a good plan, Jeremy.

jedateach:
@markguinn I've resubmitted with deprecated functions included.

markguinn:
Sorry, @jedateach I just noticed you'd resubmitted this. There's nothing stopping us from merging this is there? It all looks good to me. The only question that came up as I was having another look is whether we should keep the Order_ReceiptEmail etc classes around and mark them as deprecated. If someone were refering to those in their code that would cause a break where it won't hurt anything to leave them around and remove them entirely in 2.0.

jedateach:
@markguinn I've added in the deprecation notices to the doc blocks of the classes.

PR #308, opened 2015-02-05
Allow skipping the ContactDetails step if users are logged in, with corr...

jedateach:
...ect info already provided.


jedateach:
:+1: thanks

Issue #10, opened 2012-03-06
Enforce immutability of orders


jedateach:
Once an order is placed, it should be completely immutable, i.e. not able to change. (Apart from fulfilment related things)

Such objects:
- Order - incl totals, included items, associated member details
- Shipping and Billing Addresses
- Order Items
  - Unit Price
  - Quantity
  - Discounts / other modifications to item total
  - Total

A basic implementation could start by updating the canEdit function to check the current order status is 'Cart', or the member is Admin.

Another aspect to take into account is how the database is affected by removing DataModels, such as Modifiers, Payment types, OrderItem types, etc. Also, you should be able to import data from another ecommerce system and have order totals, and line totals remain unaffected.

The bug exists in OrderItem, where UnitPrice relies heavily on the Buyable (Product) to get it's value.
Also this commit 72c65ef85aff9f676839af2567102985aee8b16b introduces a problem where CalculatedTotal is checked in both UnitPrice and Total.

PR #309, opened 2015-02-07
Fix for a rare bug when ProductID=0 on an OrderItem

markguinn:


jedateach:
Its a bit hairy when a Product_OrderItem doesn't have a Product. I can see how it could happen when a product is deleted for example.

This fix makes sense for when this happens. Id also like to have more confidence as to what happens as data is isolated like this and db starts to get invalid. Its the whole immutability thing, which I've had some thoughts on before - see #10

markguinn:
I agree. I'm not at all sure how it happened but it did happen on one of our staging sites and then you couldn't add or remove anything from the cart. It was really confusing because it was only happening for the one browser session. Thanks for reviewing this.

PR #311, opened 2015-02-13
Update AccountPage.php for Country saved as null

AntonyThorpe:
The readonly field of Country_readonly (when operating in a single country) is not passed onto the saveaddress function.  Null is saved to the database.  Add a new textfield to the save address function.  

Is this the best solution?

### DIFF
--- a/code/account/AccountPage.php
+++ b/code/account/AccountPage.php
@@ -162,7 +162,14 @@ class AccountPage_Controller extends Page_Controller {
 		$address = new Address();
 		$form->saveInto($address);
 		$address->MemberID = $member->ID;
+
+		// Add value for Country if missing (due readonly field in form)
+		if($country = SiteConfig::current_site_config()->getSingleCountry()){
+			$address->Country = $country;
+		}
+
 		$address->write();
+
 		if(!$member->DefaultShippingAddressID){
 			$member->DefaultShippingAddressID = $address->ID;
 			$member->write();
--- a/code/checkout/CheckoutForm.php
+++ b/code/checkout/CheckoutForm.php
@@ -16,6 +16,13 @@ class CheckoutForm extends Form {
 			)
 	);
 		$validator = new CheckoutComponentValidator($this->config);
+		
+		// For single country sites, the Country field is readonly therefore no need to validate
+		if(SiteConfig::current_site_config()->getSingleCountry()){
+			$validator->removeRequiredField("ShippingAddressCheckoutComponent_Country");
+			$validator->removeRequiredField("BillingAddressCheckoutComponent_Country");
+		}
+
 		parent::__construct($controller, $name, $fields, $actions, $validator);
 		//load data from various sources
 		$this->loadDataFrom($this->config->getData(), Form::MERGE_IGNORE_FALSEISH);
--- a/code/model/Address.php
+++ b/code/model/Address.php
@@ -203,4 +203,15 @@ class Address extends DataObject{
 		$this->Company = $val;
 	}
 
+
+	function validate() {
+	    $result = parent::validate();
+	    foreach(self::$required_fields as $requirement){
+	        if(empty($this->$requirement)) {
+	            $result->error("Address Model validate function - missing required field: $requirement");
+	        }
+	    }
+    	return $result;
+	}
+
 }
--- a/tests/account/AccountPageTest.php
+++ b/tests/account/AccountPageTest.php
@@ -17,7 +17,20 @@ class AccountPageTest extends FunctionalTest{
 	}
 
 	public function testCanViewAccountPage() {
-		$this->markTestIncomplete('Log in and view account page');
+		$page = $this->get("account/");  // attempt to access the Account Page
+        $this->assertEquals(200, $page->getStatusCode(), "a page should load");
+        $this->assertContains('<form id="MemberLoginForm_LoginForm"', $page->getBody(), 'Need to login before accessing the account page');
+
+        // login using form
+        $this->submitForm("MemberLoginForm_LoginForm", "action_dologin", array(
+            'Email' => 'test@example.com',
+            'Password' => '23u90oijlJKsa'
+        ));
+
+        $page = $this->get("account/");  // try accessing the account page again
+        $this->assertEquals(200, $page->getStatusCode(), "a page should load");
+        $this->assertContains("Account", $page->getBody(), "Account Page should open");
+
 	}
 
 	public function testGlobals() {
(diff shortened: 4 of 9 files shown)


jedateach:
Thanks for picking up on this @AntonyThorpe 

To avoid setting up dummy form fields, I might instead do something like:
```php
if($country = SiteConfig::current_site_config()->getSingleCountry()){
$data['Country'] = $country;
$form->loadDataFrom($data);
}
```

Another compilmentary / similar solution might be to set a default value for "Country" on `Address`:

```yml
Address:
   defaults:
      Country: NZ
```

AntonyThorpe:
I like the first option best as the data is entered once.  
However, there is no 'Country' field in $form, only 'Country_readonly'.  Any ideas please?

jedateach:
Hmm, good point. Maybe we do need the text field. This could be a Hidden field:

```php
$form->fields->push(new HiddenField("Country"));
```

AntonyThorpe:
In Address.php, getCountryField() function, I tried:
```php
$readonly_field = new ReadonlyField("Country",
	_t('Address.COUNTRY', 'Country'),
	array_pop($countries)
);
$readonly_field->setIncludeHiddenField(true);
return $readonly_field;
```
and,
```php
$readonly_textfield = new TextField("Country",
	_t('Address.COUNTRY', 'Country'),
	array_pop($countries)
);
$readonly_textfield->setReadonly(true);
return $readonly_textfield;
```
and both record the country code correctly in the form (in the saveaddress function), however, the `$form->saveInto($address)` doesn't pass the country code to Country property in $address.

jedateach:
I see. Check the actual data values coming in, you might find they have been "namespaced". The namespacing I introduce after the form has been constructed, and it is un-namespaced on form save.

```php
var_dump($data);
```

I haven't actually checked this myself sorry.

AntonyThorpe:
Have tried to solve the missing Country field using a dynamically set default property in the address model.  This works successfully in the browser but unfortunately fails the continuous integration testing.

jedateach:
I think the problem is that it is trying to reference the SiteConfig table when it hasn't been set up yet. You'll find you get the same error if you start with a fresh/empty database.

jedateach:
We're essentially talking about defining what is an invalid address in this issue. I wonder if we should also implement the `validate()` function to ensure an address can never be written to DB in an invalid state.

Un-tested code:
```php
function validate() {
	$result = parent::validate();
	foreach(self::$required_fields as $requirement){
		if(empty($this->$requirement)) {
			$result->error("Missing required field: $requirement");
		}
	}
    return $result;
}
```
I can imagine this will break a few things, but it'll draw invalid Addresses to the surface.

---

Regarding the passing in of valid data when there is only one country, I've just realised this is actually meant to be handled inside `AddressCheckoutComponent`. See the `getData` function. Perhaps the corresponding logic for `saveData` is missing?

AntonyThorpe:
Yes you are right re the addresses in the component system, plus add the Shipping Estimate form to the list.  A single country site doesn't need a country dropdown but does need the Country to have a value to process the form and return a result.

With the checkout component system, I have come across the null Country field after adding server side validation to the form (eg max length).  A rejected form submission seems to clear the Country_Readonly field.  Haven't worked out why.  

Next steps:
populateDefaults() is a no-go.  Will cancel this. Will add the above validate() code to the Address Model and then look again at the saveaddress function.

jedateach:
@AntonyThorpe How are you with unit tests?

If you can get them set up and running, it'll quickly identify if your solution affects different parts of the system.

If you're wanting to contribute more in future, this could be a good opportunity to see how to write/update tests. In this case, you'd want to write a test that checks the Country is populated for an address when there is only one available Country. Good practice is to write the test before applying changes. i.e. make sure the test picks up the issue as a failure, then write the code to fix it.

AntonyThorpe:
Thank you for your advice and encouragement.  Have created some new commits following writing/updating tests.  However, CheckoutComponentTest fails.  I don't understand this test.  Can you please point me in the right direction?  What model logic is it trying to test?

markguinn:
@AntonyThorpe it looks to me like CheckoutComponentTest is useless right now anyway. It looks like just a placeholder. It's something I've been wanting look into and improve, actually. I have a PR (https://github.com/burnbright/silverstripe-shop/pull/266) that I'm waiting to merge until I have a chance to improve those tests a bit. I'd suggest looking to add a test method to either CheckoutTest or CheckoutForm test that would cover your addition (I'm not sure about the best place - I'll leave that up to your judgment). As long as it covers the specific case of the PR I'd be happy to go ahead and merge this. Thanks for tackling this!

markguinn:
(also, note that we're having some trouble with our Scrutinizer setup at the minute so you only need to pass Travis)

AntonyThorpe:
Any feedback please regarding the latest commits or this pull request overall?
Latest commits: 
- removed the need to validate the Country fields for single country sites in CheckoutForm.php (as it is not entered in the form) and updated CheckoutFormTest as previously recommended;
- added an additional assertion to AccountPageTest; and,
- added a suite of tests to CheckoutComponentTest.  This no longer fails as previously advised (have nutted it out).
Thanks

markguinn:
Don't worry about Scrutinizer right now. We're having some trouble with it and looking into alternatives. I'm not sure why the Travis build is failing. The code looks good to me. (Also, thank you for this contribution and the excellent test coverage)

markguinn:
Ha, just realized I'm sounding very repetative about that Scrutinizer thing. Oops. I just re-ran the Travis build and the Mysql run now is passing. I'm re-running the postgre as well. Could have just been a fluke in their system. Beyond that, this all looks good to me. Are you happy with this @jedateach ?

markguinn:
It looks like there may be a legitimate failure case with Postgre.

AntonyThorpe:
Hi
Just added a commit to the "SaveCountryAccountPage" branch.  One failure with Travis with the SQLite DB (all others passed).  See below:
Problem 1
- burnbright/silverstripe-listsorter 2.0.0 requires silverstripe/cms 3.* -> satisfiable by silverstripe/cms[3.0.x-dev, 3.1.x-dev, 3.2.x-dev, 3.x-dev, 3.3.x-dev].
- burnbright/silverstripe-listsorter 2.0.1 requires silverstripe/cms 3.* -> satisfiable by silverstripe/cms[3.0.x-dev, 3.1.x-dev, 3.2.x-dev, 3.x-dev, 3.3.x-dev].
- Can only install one of: silverstripe/cms[4.0.x-dev, 3.1.x-dev].
- Can only install one of: silverstripe/cms[4.0.x-dev, 3.2.x-dev].
- Can only install one of: silverstripe/cms[4.0.x-dev, 3.x-dev].
- Can only install one of: silverstripe/cms[4.0.x-dev, 3.3.x-dev].
- Can only install one of: silverstripe/cms[4.0.x-dev, 3.0.x-dev].
- Installation request for silverstripe/cms 4.0.x-dev -> satisfiable by silverstripe/cms[4.0.x-dev].
- Installation request for burnbright/silverstripe-listsorter ~2.0 -> satisfiable by burnbright/silverstripe-listsorter[2.0.0, 2.0.1].

Do we need to update SS Shop's composer.json file?  The listsorter's latest version is 2.0.1.

Scrutinizer also errored.  It's report says:
"The code coverage data was not received within the specified time. Please make sure your third-party service is configured to send code coverage data."
Was it waiting for a report from Travis?

Any ideas please?

Thanks Antony

markguinn:
Brilliant! That last build is allowed to fail. We're still working on getting compatible with master branch. We've also solved our Scrutizer problems. If you rebase onto the latest master and squash these commits into one we can merge this. If you're not comfortable with the git gymnastics involved I can walk you through that or just take care of it for you. Thanks for your work on this!

AntonyThorpe:
Bingo - squashed!  Thank you Mark.

Issue #312, opened 2015-02-17
Make AddProductForm more extensible


jameshoward:
If I'm right, SilverStripe’s `Form` class doesn't have a nice way of pushing new fields into an already-constructed instance. You can’t even do `$this->fields->push($newField)` without first calling `setForm` on the new field.

To make it easier to extend `AddProductForm`, it could be refactored to build its fields, actions and validator in separate protected methods `getFields`, `getActions` and `getValidator`. This would allow an overriding class to do as so:

```php
protected function getFields() {
    $fields = parent::getFields();
    $fields->push(new TextField('Test', 'Test'));
    return $fields;
}
```

This would make for cleaner inheritance without the need to copy/paste/edit the `__construct` method.

jedateach:
@jameshoward I'm pretty sure you *can* update the fields of a form that has been instantiated already.

To access Form's fields, call `$form->Fields()`
 Why does form not have `getFields`? I'm not sure.

jedateach:
@jameshoward In all my time working with SilverStripe I've not really had issues setting fields on an already constructed instance, and **not** setting `$field->setForm($form)`. I'm not sure why this is - maybe `setForm` is actually called somewhere else.

But regardless, I think this refactoring tidies things up a bit. I've often put things in the constructor, and assumed that an `Extension` would be used to update the fields. But your approach just opens up even more flexibility.

jameshoward:
@jedateach Yeah, I was doing `$this->fields` from a subclass, which was getting what I expected, a `FieldList`. However, adding fields with `push` caused a non-object error because of https://github.com/silverstripe/silverstripe-framework/blob/3.1/forms/FormField.php#L167

It was trying to get the `Form` related to my added `FormField` and failing. Maybe I'm doing something wrong but I recall having a similar issue previously with a similar attempt to add fields to an existing form rather than building up a field list in the constructor.

jedateach:
I see - thanks for the explanation.

PR #313, opened 2015-02-17
Refactor AddProductForm for greater extensibility.

jameshoward:
Split fields, actions and validator creation into protected methods that can be overriden in sub-classes.

Relates to #312


jedateach:
Thanks @jameshoward - welcome to the shop contributors list! :)

PR #319, opened 2015-03-16
Handle missing required values consistently in MatchObjectFilter.

jameshoward:
There is a discrepancy between how has_one and db fields are handled when required but no matching data is supplied.

The former results in a filter of `("FieldName" = 0 OR "FieldName" = NULL)` while the latter produces `"FieldName" = null`.


jameshoward:
The failing test on PostgreSQL made me realise that the = 0 check didn't make sense within the context of a database field, only on relations. Removing then when dealing the a field in `db()`.

jameshoward:
Looks from the latest CI build that this would have a negative effect on some other tests. Will close this PR for now and re-open one once all tests are passing.

PR #320, opened 2015-03-19
Update Product.php for character length of Model

AntonyThorpe:
Need a little more space please for our Model names


jedateach:
Ultimately, I think I'd rather remove this field from the shop core, as it doesn't necessarily apply to all applications of the shop.

You can actually add/edit fields in by updating the db array in yaml config:
```
Product:
    db:
       Model: 'Varchar(50)'
```

PR #321, opened 2015-03-24
Handle missing required values consistently in MatchObjectFilter.

jameshoward:
There is a discrepancy between how `has_one` and `db` fields are handled when required but no matching data is supplied. The former results in a filter of `("FieldNameID" = 0 OR "FieldNameID" = NULL)` while the latter produces `"FieldName" = null` (invalid).

The PR changes the behaviour so that `db` fields result in a filter of `"FieldName" IS NULL`.

Also updated existing `CustomProductTest` and created `MatchObjectFilterTest`.

### DIFF
--- a/code/cart/MatchObjectFilter.php
+++ b/code/cart/MatchObjectFilter.php
@@ -77,10 +77,13 @@ class MatchObjectFilter{
 		foreach($fields as $field => $value){
 			$field = Convert::raw2sql($field);
 			if(array_key_exists($field, $db)){
-				$dbfield = $singleton->dbObject($field);
-				$value = (isset($this->data[$field])) ? $this->data[$field] : null;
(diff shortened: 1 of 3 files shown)


markguinn:
This looks good to me, @jameshoward. Thank you for your work on this! Would you mind squashing these commits (http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) so we can merge this?

jameshoward:
@markguinn No problem. Do you want them squashing into a single commit?

markguinn:
That would be great.

jameshoward:
@markguinn If that worked as expected that should now be a single commit documenting all the changes.

markguinn:
Great! Thanks again for your work on this. Good catch on this bug.

PR #323, opened 2015-03-27
Add canView check to Order.

### LINKED ISSUE #322: Only admins can view orders by default.
There is no `canView` defined in `Order` so nobody but users in the Administrator group can view orders.

The simplest solution would be to add a `canView` that returns `true` in the same way as `canEdit` does.

jameshoward:
Without an implementation of canView, only administrators could view
order details.

Fixes #322

### DIFF
--- a/code/model/Order.php
+++ b/code/model/Order.php
@@ -341,6 +341,14 @@ class Order extends DataObject {
 	}
 
 	/**
+	 * Check if an order can be viewed.
+	 * @return boolean
+	 */
+	public function canView($member = null) {
+		return true;
+	}
+
+	/**
 	 * Check if an order can be edited.
 	 * @return boolean
 	 */


markguinn:
I agree that we need to add support for this to the shop core, but I wonder if we need to be more robust and actually add a permission for being able to view/edit/etc an Order. I've always done the same as this in our projects using an extension and just given users access to the Orders ModelAdmin in the CMS. That's as far as any of our sites have ever needed and may be all that's required unless one is using a REST API for Orders or something. 

@jedateach and @wilr do either of you have more experience with more granular permissions in SS and care to comment on this? I think we should take the simplest route and not over-complicate this.

markguinn:
(also, thanks for noticing this and working up a PR, @jameshoward!)

jameshoward:
If it wasn't for `canEdit` leading the way I would have suggested a new permission or re-use of the `CMS_ACCESS_OrdersAdmin` that ModelAdmin will have created.

jameshoward:
Also worth noting that the same is true of `Payment`. I can now view orders but the payments show as empty unless an admin. The whole lot could work under a single permission such as CMS_Shop.

jameshoward:
(@markguinn no worries, glad to be able to contribute to something we're making good use of)

markguinn:
Interesting. I hadn't realized canEdit was already present. As such I'd be comfortable with merging this but hopefully adding more complete permissions in 1.1 or 2.0 (seems like a border-line backwards incompatible change to me if we're suddenly requiring an additional permissions for users that previously could see an order). I'd like to hear Jeremy's input though since that's a longer term question.

jameshoward:
Yes, good point on it potentially being a breaking change.

PR #324, opened 2015-04-17
Added requireDefaultRecords to AccountPage

RVXD:
Added requireDefaultRecords function to AccountPage as it is needed to use the checkout.

### DIFF
--- a/code/account/AccountPage.php
+++ b/code/account/AccountPage.php
@@ -54,7 +54,7 @@ class AccountPage extends Page {
 			$page->write();
 			$page->publish('Stage', 'Live');
 			$page->flushCache();
-			DB::alteration_message('Checkout page created', 'created');
+			DB::alteration_message('Account page created', 'created');
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


markguinn:
Thanks for the pull request, @RVXD. One small thing - would you mind squashing these commits into one so we can keep the commit log clear and easier to make sense of? (https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/ - you can just force push to your same branch to update this PR)

On a bigger picture note, I wonder if the config setting here should be moved to ShopConfig and defaulted to true. That would be a backwards-incompatible change so would have to wait for 2.0 unless we checked for both values in which case it could be a 1.1 thing. At minimum I'd say we ought to add these flags to example_config.yml so they can get some use. @jedateach do you have any thoughts about that? What about you, @RVXD as someone who's obviously using this feature?

markguinn:
Sorry if I confused the issue here. I'm happy to merge this PR once it's squashed and we can deal with the configuration issues in another PR.

RVXD:
Thanks for your reply. I'll try to 'squash' the commits into one and request a pull again.
The config setting is also in other requireDefaultRecords, that's why I copied it in.

markguinn:
I squashed these and merged this manually. Welcome to the committers list, @RVXD

PR #325, opened 2015-04-18
Testing some different CI configurations

markguinn:


markguinn:
@jedateach you can ignore this PR. I'm just trying to figure out the travis/scrutinizer issues we're having.

markguinn:
I just emailed their support and cc'ed you. Hopefully they'll have some insight. It's got to be discouraging for contributors or people checking out the project to see all those red x's in the PR's.

jedateach:
Great, thanks for following that up with the Scruitinizer team.

markguinn:
@jedateach I've never heard anything back from Scrutinizer support on this. Would you be at all open to switching to a competitor such as Code Climate (https://codeclimate.com/github/markguinn/silverstripe-shop) which also can do code coverage and seems to work similarly. From what I've read Scrutinizer is better, but I also think it may be best to just get something working and move on to more important matters like Behat tests. Thoughts?

markguinn:
This would also be an option - https://insight.sensiolabs.com/projects/22c5135b-a33c-452a-b6e5-f5b8cccf5c74/analyses/1

I've heard it's more powerful and therefore more helpful but it also seems to be a bit more opinionated about things that are part of Silverstripe's style (multiple classes to a file, camel-casing method names, etc).

markguinn:
Interestingly, it looks like we're back to only failing on the code coverage so I might spend a few more minutes on that and hopefully can stick with Scrutinizer.

markguinn:
Closing this because Travis seems to have started ignoring it.

PR #327, opened 2015-04-18
Allow extending Product Attribute Type with updateCMSFields

webtorque7:


markguinn:
Thanks for contributing! One little thing - it looks like you've got spaces instead of tabs there. Would you mind fixing that so we can keep the codebase tidy? You'll need to squash commits down to a single one (https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/) and then do a force push to this branch.

webtorque7:
My bad, still had it on psr format, will resubmit

webtorque7:
Ok, should be cleaned up now

PR #329, opened 2015-04-20
Created simple .editorconfig

markguinn:
For contributors who's editors/IDE support it, this will save us from having to push back on some style issues in PR's (primarily spaces => tabs). Check out http://editorconfig.org/ for more information.

I'll leave this open for a week or so for anyone to push back and/or add rules to the file as you see fit.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# For more information about the properties used in
+# this file, please see the EditorConfig documentation:
+# http://editorconfig.org/
+
+[*]
+charset = utf-8
+indent_style = tab
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.yml]
+indent_size = 2
+indent_style = space
+
+[*.json]
+index_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,6 +5,9 @@
 [*]
 charset = utf-8
 indent_style = tab
+end_of_line = lf
+insert_final_newline = true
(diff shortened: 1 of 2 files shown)


markguinn:
Here are a few other examples from PHP projects. 

* CakePHP - https://github.com/cakephp/cakephp/blob/master/.editorconfig
* Drupal - http://cgit.drupalcode.org/drupal/tree/.editorconfig
* Symfony2 - https://github.com/symfony/symfony/blob/2.7/.editorconfig

Here's the full list: https://github.com/editorconfig/editorconfig/wiki/Projects-Using-EditorConfig

jedateach:
Nice. Looks like a good initiative to get behind.
Any talk of this in SilverStripe community already?

markguinn:
I've not seen it showing up on any other Silverstripe projects, no. I suppose one step beyond this is getting Scrutinizer to also check for style issues (where braces are placed, etc). The nice thing about this is that it's proactive if your editor supports it. So I could use PSR-2 on the rest of a project but any file in the shop folder will automatically use tabs, etc. 

I'd love to see it adopted more widely within the SS ecosystem. I wonder if it's worth raising on the dev list?

Note to self: before merging this we should also update the contributors doc asking them to install editor/ide plugins as needed.

markguinn:
@jedateach are you happy with the state of this? If so I'll go ahead and squash these commits and merge.

jedateach:
Yep, great. Go for it!

PR #330, opened 2015-04-20
Refactored CartForm so that code can be reused in other contexts, and re...

jedateach:
...structured code to be a bit tidier

### DIFF
--- a/code/cart/CartForm.php
+++ b/code/cart/CartForm.php
@@ -10,20 +10,28 @@ class CartForm extends Form{
 
 	public function __construct($controller, $name = "CartForm", $cart = null, $template = "Cart") {
 		$this->cart = $cart;
-		parent::__construct($controller, $name, new FieldList(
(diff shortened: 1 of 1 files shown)


code/cart/CartForm.php
```diff
@@ -10,20 +10,28 @@ class CartForm extends Form{
...
 	}
 
-	public function editableItems($items) {
+	/**
+	 * Add quantity, variation and remove fields to the
+	 * item set.
+	 * This method is static so that it can be used externally.
+	 * @param SS_List $items
+	 */
+	public static function addEditFields(SS_List $items) {
```
wilr: Should statics follow the SilverStripe convention of add_edit_fields?

jedateach: I've updated the function to fit the convention.

jedateach:
Cancel that, I'm going to make some further changes.

PR #332, opened 2015-04-23
Clean up

jedateach:
Tidied code.

* Tidied ViewableCart to use new ORM, and removed unused property
* Tidied MatchObjectFilter ->getFilter to be a bit more readable.
* FIX: Order->debug was showing the items and modifiers lists twice.

### DIFF
--- a/code/cart/MatchObjectFilter.php
+++ b/code/cart/MatchObjectFilter.php
@@ -54,7 +54,6 @@ class MatchObjectFilter{
 	 * @return array of filter statements
 	 */
 	public function getFilter() {
-
 		if(!is_array($this->data)){
 			return null;
 		}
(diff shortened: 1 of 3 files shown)


markguinn:
This all looks good to me, @jedateach.

jedateach:
@markguinn I'm guessing you're ok with me doing the merge then?
I'm kind of thinking a good practice is that the person who wrote the code should not be the one who merges it. 
In this regard, I need #333 looked at for a current project.

markguinn:
No problem, Jeremy. I thought since you hadn't squashed the commits you might be planning to add more to it. Since this is really 3 changes I'd be ok with leaving them as three commits if you are.

PR #340, opened 2015-05-13
Fixes UX issue when an extension creates a price at higher precision …

markguinn:
…than the currency supports (e.g. 3.145)

### DIFF
--- a/code/product/Product.php
+++ b/code/product/Product.php
@@ -223,7 +223,7 @@ class Product extends Page implements Buyable {
 				try {
 					if($variation->canPurchase($member, $quantity)) {
 						$allowpurchase = true;
-					
+
 						break;
 					}
 				} catch(ShopBuyableException $e) {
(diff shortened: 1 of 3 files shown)


jedateach:
My concern with this change is that we're immediately blocking out hopes of using bitcoin as a currency, rather than making inroads to use it, and currencies like it.

Note that Order has a $rounding_precision static variable, which is used in some of it's output.

I understand there a fix needed for an immediate issue. Perhaps you could just adjust your PR to allow the precision to be customisable?

I can't think off the top of my head the best way to solve / clean up this stuff, but I think it does need some planning before implementing a final once-for-all solution.

markguinn:
I agree. I was hoping you'd have some ideas here. I didn't know about Order::$rounding_precision but that seems like a good start to me. I'm a little surprised at how limited the SS Currency class is. I suppose options would be:

1. Subclass Currency and let it do the rounding in setValue (which would allow anyone to use the Injector to swap out a BitCoinCurrency class or something)
2. Transition to using Money which would be a breaking change
3. Stick with $rounding_precision for now

This is not urgent for us because I'm using a custom branch for the project in question and merging in patches as needed. I'd rather get the right solution than push something through.

jedateach:
The module actually already does subclass Currency, and replaces it - which is a bit of a hack, and would never allow anyone else to do the same.

The thing I don't like about the Money class is that there is always a currency associated with it. It is not easy to pass around variables. I don't think there is need to be working in multiple currencies when storing/calculating things. It is only once you reach the final point of displaying values that you want values to be converted and formatted. I suppose an Order should have an associated currency. Or perhaps just the payment for that order?

Yes I agree re: getting things right, rather than hastily pushing something through. I think it'd be ok for now to rely on $rounding_precision, as it'll fix the immediate problem and *hopefully* allow for customisation, if it is wanted.

I think ultimately Currency and Money should be reviewed on a larger SilverStripe core scale to determine where things are lacking, and how to remedy. I recall there are bits of discussion around, and I think it somewhat relates to using Zend libraries, which I've found to be quite big/scary/hairy, but probably worth considering amongst other things.

markguinn:
@jedateach I've updated this to respect rounding_precision and added a unit test to cover it. Are you happy to go ahead and merge?

PR #341, opened 2015-05-22
Moved the CheckoutPage extistingValues toggle code

jedateach:
… outside of $(document).ready

This prevents fields disappearing after page has fully loaded, which could be seconds later if other javascript is loading/running (especially external scripts). I think we can assume the relevant DOM will be loaded since javascript is added to the bottom of the page.

I've also reorganised code to be  narrower, thus readable.


markguinn:
This looks good to me.

PR #345, opened 2015-06-13
Tentatively CI should be fixed

markguinn:


markguinn:
This is just super frustrating. Travis is so fragile right now, mostly because composer keeps timing out. It looks like we're not the only ones having trouble though. Even the silverstripe-framework repository is having the same timeout issues we're having. I've lodged an issue with the silverstripe travis support module (https://github.com/silverstripe-labs/silverstripe-travis-support/issues/15). 

I'm out of time to work on this for now so I suppose the best approach is wait and see if it's an infrastructure issue that will resolve itself.

PR #348, opened 2015-06-23
Trying to test against silverstripe master

markguinn:


markguinn:
I'm going to go ahead and close this out. The module works fine under SS4.0 but testing is still difficult because of other module dependencies and some strange quirks in the mock controller routing in CheckoutTest. I think our time is better spent elsewhere until the 4.0 release gets closer.

PR #349, opened 2015-06-24
Update AccountPage.php for a form extension point

AntonyThorpe:
Provide an extension point in the CreateAddressForm in AccountPage.php.  Thanks.

### DIFF
--- a/code/account/AccountPage.php
+++ b/code/account/AccountPage.php
@@ -199,6 +199,12 @@ class AccountPage_Controller extends Page_Controller {
 	public function ChangePasswordForm() {
 		$form = new ChangePasswordForm($this, "ChangePasswordForm");
 		$this->extend('updateChangePasswordForm', $form);
+		$this->data()->extend('updateChangePasswordForm', $form);
+
+		if($this->data()->hasMethod('updateChangePasswordForm')){  // if accessing through the model
+			Deprecation::notice('2.0', 'Please access updateChangePasswordForm through AccountPage_Controller instead of AccountPage (this extension point is due to be removed)');
+		}
+
 		return $form;
 	}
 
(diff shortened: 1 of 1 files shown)


markguinn:
Should the extension point for this not be on AccountPage_Controller? I think in SS forms are usually tied to controllers instead of models.

AntonyThorpe:
Good point.  ChangePasswordForm in AccountPage's controller also references the model.  Should I update both?

markguinn:
Ugh. That could potentially trip people up and break their code if they were already using it. Does it seem to excessive to you to allow extension in both places? That code is not going to be called very often so I don't see a speed issue. Something like the following in both places would present a more sensible api and not break backward compatibility:

```
$form = new Form($this, "CreateAddressForm", $fields, $actions, $validator);
$this->data()->extend('updateCreateAddressForm', $form);
$this->extend('updateCreateAddressForm', $form);
return $form;
```

AntonyThorpe:
I don't think we need the extension point on both the model & controller.  Looking at the history, it was my commit that created the updateChangePasswordForm extension on the model last November.  As this is so recent, I think there is minimal risk of breaking other people's code.  The most recent commit places both updateCreateAddressForm and updateChangePasswordForm extension points on AccountPage_Controller.  Please advise.  Thanks.

markguinn:
It looks like your change was incorporated into 1.0.0, which means we'd lose backward compatibility and I feel pretty strongly about maintaining semantic versioning in our releases. As I've thought more about it I'd prefer to allow the extension on both the model and controller for now, with a comment that extension point on the model is deprecated will probably disappear in 2.0.

On a related note, do you have any idea why that test just started failing in postgre? That same test was failing on your other PR, but then stopped. I can't see any reason it would fail unless there was some caching thing specific to postgre. I won't stop that from merging this PR but I'll want to look into it later to make sure we didn't introduce a bug.

AntonyThorpe:
Updated the pull request as advised to preserve backward compatibility and included a deprecation notice.
Hey, all tests have passed including the ones in postgre.  Sorry, I am not sure why these failed last time.

markguinn:
Perfect! Thank you. We can just keep an eye on that test. I suspect if anything it's a bug in the test itself and/or the travis config so I don't expect it will hurt any actual sites.

PR #351, opened 2015-07-05
Trying to fix a flaky test

markguinn:


tests/account/AccountPageTest.php
```diff
@@ -119,7 +119,7 @@ public function testAddressBookWithReadonlyFieldForCountry() {
  		$this->submitForm("Form_CreateAddressForm", "action_saveaddress", $data);
  		$this->assertEquals(200, $page->getStatusCode(), "a page should load");
 
- 		$nz_address = Address::get()->last();
+ 		$nz_address = Address::get()->filter('PostalCode', '8011')->sort('ID')->last();
```
jedateach: Yep, good to lock this down a bit tighter. 
:+1:

markguinn: Yeah, I'm not sure what was happening here but that's the only reason I can think of that the test would fail randomly and only ever on PostGre.

PR #352, opened 2015-07-06
Update en.yml

hex0id:
Postal Code string had a wrong positioning, so the label for BillingAddress in forms was "Billing Address Postal Code"


markguinn:
Thanks! Welcome to the contributors list. :)

PR #362, opened 2015-09-11
Add extension point for ProductsShowable

hailwood:
Provides an extension point so you can apply extra filters . e.g. brands, cost range etc.


hailwood:
As an example of the usage for this

```php
<?php

class ProductCategoryExtension extends SiteTreeExtension {

	/**
	 * @param \DataList|Product[] $products
	 */
	public function updateProductsShowable(DataList &$products) {
		if(Session::get('BrandFilter')){
			$products = $products->filter('BrandID', Session::get('BrandFilter'));
		}
	}

}
````

PR #363, opened 2015-09-11
Added CheckoutStep_Summary component config extension hook.

jedateach:
+Minor bugfix and tidy


jedateach:
should have rebased first...

PR #365, opened 2015-09-11
Added CheckoutStep_Summary component config extension hook.

jedateach:
+Minor bugfix and tidy

Allow summary checkout component config to be updated in extensions./

Fix to prevent payment field manipulation in the case that such fields don't exist

### DIFF
--- a/code/checkout/CheckoutPage.php
+++ b/code/checkout/CheckoutPage.php
@@ -117,7 +117,7 @@ class CheckoutPage_Controller extends Page_Controller {
 		}
 
 		$config = new CheckoutComponentConfig(ShoppingCart::curr(), false);
-		$config->AddComponent(new OnsitePaymentCheckoutComponent());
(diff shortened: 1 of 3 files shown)


markguinn:
@jedateach I just noticed that both of these are against master instead of develop which breaks our gitflow workflow. I'm going to try to sort them out and merge them into develop manually.

jedateach:
Oh right - I wasn't exactly sure which to push to, but since I'm working on a branch closer to master, I just went for the easy option.

Could the 'Contributing' docs be updated to give an explanation of the branches, which branch to commit to etc?
I understand that you wouldn't want to support or add features to older release branches. Perhaps only point fixes. But yeah, all stuff that could do with some docs to make it easer for contributors. :)

markguinn:
That's a good idea. It does mention that we use gitflow but it would be good to spell out which branch to PR against.

PR #367, opened 2015-09-20
Changed range value to be SellingPrice not Price

mediabeastnz:
Changed the range value to be SellingPrice not Price on Product Variations.

If variations were extended via updateSellingPrice() then this function would of just used the standard Price. This will prevent that.

### DIFF
--- a/code/product/variations/ProductVariationsExtension.php
+++ b/code/product/variations/ProductVariationsExtension.php
@@ -46,7 +46,7 @@ class ProductVariationsExtension extends DataExtension {
 		if(!$variations->exists() || !$variations->Count()){
 			return null;
 		}
-		$prices = $variations->map('ID','Price')->toArray();
+		$prices = $variations->map('ID','SellingPrice')->toArray();
 		$pricedata = array(
 			'HasRange' => false,
 			'Max' => ShopCurrency::create(),


mediabeastnz:
How can i fix that scrutiniser issue? Is that a code issue or service?

markguinn:
It's not your fault. I had it fixed and it seems to have

markguinn:
Am I understanding this correctly that mapping "SellingPrice" instead of "Price" causes the sellingPrice method to be called which in turn applies discounts, etc? This seems like a good thing to me.

I just checked and this method is not covered by any unit tests. Would you have time and expertise to write a quick test for this method? If not I'm happy to merge this as-is.

mediabeastnz:
I see in ProductVariationTest.php theres this check `$this->assertEquals(22, $variation->sellingPrice(), "Variation price is $22 (price of ball");`

Would that be enough to check? If not let me know what the test should do and i'll happily try and write one :+1:

markguinn:
Sure, testing the PriceRange method on Product (via ProductVarationsExtension) would involve setting up a product with a few variations (probably already done in the fixtures) and checking that this method correctly returns the range from the cheapest variation to the most expensive.

markguinn:
@mediabeastnz - I'm going to release 1.1.1 or 1.2.0 in the next week or so. Are you still hoping to write a test for this? If you don't think you'll have time soon, I might go ahead and merge this so it makes the next release.

mediabeastnz:
Sorry I probably won't have time! I haven't had to much experience in writing tests anyways. If someone else could do one that'd be awesome so I can see how it's done :+1:

markguinn:
Sure, no problem. I'll go ahead and merge this and then push a test in a later commit. Thanks for contributing!

markguinn:
@mediabeastnz - if you're interested in what the test for this looks like: https://github.com/burnbright/silverstripe-shop/blob/0eda8099c1d40048285ae61575a4367e018fa236/tests/products/variations/ProductVariationTest.php#L75

Issue #369, opened 2015-09-25
Address doesn't use customized required_fields in validate method.


bummzack:
**Premise:** `Address::required_fields` should be customizable via config files (or in code by updating the config value).

This seems to be nicely implemented in `getRequiredFields`, but the `validate` method doesn't use these required fields, it reads directly from the static variable, which doesn't allow customization.

**Proposed fix:** Use `getRequiredFields` in the `validate` method as well.

Can create a PR for this if wanted.

markguinn:
That sounds like a holdover from SS2.4. Good idea. If you've got time to create that PR that would be helpful. Thanks.

bummzack:
I just noticed that this has been changed in the `develop` branch (I was looking at the 1.1.0 version). The `validate` method now uses `$this->config()->required_fields`.

I think it would still be better to use `getRequiredFields` instead, since it handles overrides and extensions as well.

markguinn:
I agree

bummzack:
Created a PR for this. #370

markguinn:
Solved in #370

PR #370, opened 2015-09-28
Fix issue #369. Use `getRequiredFields` in the Address' validate method.

bummzack:
Added UnitTest to ensure functionality.

### DIFF
--- a/code/model/Address.php
+++ b/code/model/Address.php
@@ -207,12 +207,12 @@ class Address extends DataObject{
 	function validate() {
 	    $result = parent::validate();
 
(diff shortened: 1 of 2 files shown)


markguinn:
@bummzack did the tests pass for you locally? If so I'll go ahead and merge this. I need to spend some time on our CI setup very soon! I don't think any of the failures above are the result of this PR.

Also, THANK YOU for submitting tests with this! Brilliant.

bummzack:
@markguinn Yeah, the Test passed locally. Outdated composer installs seem to be the reason for a lot of CI fails :)

PR #373, opened 2015-10-05
Adding back some changes from previous reverted commits

markguinn:
These two commits broke the build and may have broken checkout with SagePay. Placing them here so we can get Travis passing again and think through the consequences of changing PaymentForm to CheckoutForm.

### DIFF
--- a/code/checkout/CheckoutForm.php
+++ b/code/checkout/CheckoutForm.php
@@ -1,20 +1,27 @@
 <?php
 
+/**
+ * @package shop
+ */
 class CheckoutForm extends Form {
 
 	protected $config;
 	
 	protected $redirectlink;
 
+	private $checkout_action_text = 'Proceed to payment';
+
 	public function __construct($controller, $name, CheckoutComponentConfig $config) {
 		$this->config = $config;
 		$fields = $config->getFormFields();
+
 		$actions = new FieldList(
 			FormAction::create(
 				'checkoutSubmit',
-				_t('CheckoutForm.PROCEED', 'Proceed to payment')
+				_t('CheckoutForm.PROCEED', $this->config()->get('checkout_action_text'))
 			)
-	);
+		);
+
 		$validator = new CheckoutComponentValidator($this->config);
 		
 		// For single country sites, the Country field is readonly therefore no need to validate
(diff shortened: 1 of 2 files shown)


wilr:
@markguinn Taking a look at the travis failure, is it just simply a fact AccountPage has not been setup in the test? That probably should not be an assumption. Regarding the CheckoutForm to PaymentForm, I'd have thought the PaymentForm should only be used on Payment action, not Checkout?

markguinn:
Yes, I think the test should be simple to fix. 

I don't feel I have a clear handle on how CheckoutForm and PaymentForm relate, to be honest, especially once you add in single-page checkout vs multistep checkout, and some payment processors process inline before the request finishes while others redirect to an external URL. I was hoping slievr would have replied on gitter by now about whether that was the source of his problems on SagePay. @jedateach do you have any insights on how PaymentForm and CheckoutForm relate and whether the above changes would cause problems for offsite processors?

In a few months we'll hopefully have this stuff covered by the behat tests but we're just not quite there yet.

code/checkout/CheckoutForm.php
```diff
@@ -1,20 +1,27 @@
...
+
 	public function __construct($controller, $name, CheckoutComponentConfig $config) {
 		$this->config = $config;
 		$fields = $config->getFormFields();
+
 		$actions = new FieldList(
 			FormAction::create(
 				'checkoutSubmit',
-				_t('CheckoutForm.PROCEED', 'Proceed to payment')
+				_t('CheckoutForm.PROCEED', $this->config()->get('checkout_action_text'))
```
jedateach: I make this text update using a lang file.

jedateach:
So PaymentForm extends CheckoutForm. CheckoutForm handles the setup and submission of checkout component based forms.

PaymentForm simply adds the final processing step of a checkout steps flow. It is not perhaps clear that the PaymentForm is needed for the final submission, whether payment is required or not.

I think there is nothing particularly fancy about their separation, other than making the final processing logic __only__ available to the final step in the checkout steps.

code/checkout/CheckoutPage.php
```diff
@@ -86,7 +89,7 @@ public function OrderForm() {
 			return false;
 		}
 
-		$form = PaymentForm::create(
+		$form = CheckoutForm::create(
```
jedateach: If this is causing PaymentForm to not be available during the final steps / payment handing, then I suspect that this is where the SagePay issue may be.

code/checkout/CheckoutForm.php
markguinn: I agree on this one. I think it reduces complexity not to have too many config variables.

markguinn:
@wilr were you encountering a problem that prompted this change? Are you using a single page checkout?

wilr:
Correct, single page without a payment step.

PR #376, opened 2015-10-20
Fix #375 template syntax

### LINKED ISSUE #375: Update template syntax for translations and variables.
This is a non-breaking issue that should be done prior to tackling issue #368.

Here's what should be done:

**Use `$Variable` in templates instead of `Variable`.**
Before: `<% if Cart %>`
After: `<% if $Cart %>`

**Use the proper syntax for translations.**
Before: `<% _t("PRODUCT","Product") %>`
After: `<%t PRODUCT "Product" %>`

This change should not change any existing behavior and will ensure that #368 doesn't become too large of a PR.

bummzack:
Updated variable-syntax in templates from `Variable` to `$Variable`.

### DIFF
--- a/templates/Includes/AccountNavigation.ss
+++ b/templates/Includes/AccountNavigation.ss
@@ -13,13 +13,13 @@
             </li>
 
             <li>
-                <a href="{$Link(editprofile)}">
+                <a href="{$Link('editprofile')}">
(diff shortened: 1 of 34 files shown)


anselmdk:
Looks good to me. To my knowledge your changes should wok on any SilverStripe install since `3.0` onwards.

bummzack:
I just noticed that I introduced a bug by using `$LinkingMode` instead of `$LinkOrSection`. Will create a new PR.

PR #377, opened 2015-10-20
Fix for #375: Updated template syntax

bummzack:


markguinn:
I feel like I spend all of my time getting Travis and Scrutinizer to talk. So frustrating. Anyway, thanks for this!

PR #378, opened 2015-10-22
Converted all PHP language files to YML and merged with existing YML files.

bummzack:
Merged files where YML and PHP versions existed.
Deleted Backup-Files.
Ran all YML files through validator (http://www.yamllint.com/). (+1 squashed commit)
Squashed commits:
[16f6bc9] Renamed existing YML files by appending `.bak` to them.
Converted PHP language files to YML (using https://github.com/chillu/i18n_yml_converter).


markguinn:
If I understand correctly, there's very little chance of this breaking anyone's site, correct? The php versions haven't really worked since SS2.4?

bummzack:
Yes, this shouldn't break anything. Firstly, PHP files didn't work since 2.4 and even if somebody used the legacy-adapter, the YML versions are 1:1 conversions from the PHP files. So all translation keys/values will be identical to PHP. 

This is pretty much a drop-in replacement.

Issue #380, opened 2015-10-28
SS 3.2  User Error while creating new product...


Edvardas-idea:
[User Error] Uncaught SS_DatabaseException: Couldn't run query: SELECT DISTINCT ...
The problem is in this line somewhere:
 https://github.com/burnbright/silverstripe-shop/blob/develop/code/product/Product.php#L111 , i think it should be easy to fix :)

PR #385, opened 2015-11-02
Fixed issue #380

StephenHonor:
Added a new function to retrieve category options without ancestors
included in the list.

The issue was caused by the main category not yet being selected

### DIFF
--- a/code/product/Product.php
+++ b/code/product/Product.php
@@ -104,12 +104,10 @@ class Product extends Page implements Buyable {
 		//general fields
 		$fields->addFieldsToTab('Root.Main', array(
 			TextField::create('InternalItemID', _t('Product.CODE', 'Product Code/SKU'), '', 30),
-			DropdownField::create('ParentID', _t("Product.CATEGORY", "Category"), $this->categoryoptions())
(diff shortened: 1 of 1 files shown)


code/product/Product.php
```diff
@@ -170,6 +168,18 @@ private function categoryoptions() {
 	}
 
 	/**
+	 * Helper function for generating a list of additional categories excluding the main parent.
+	 * @return array categories
+	 */
+	private function categoryoptionsnoparent() {
```
markguinn: This is great. Could you rename this method to getCategoryOptionsNoParent to match up with SilverStripe (and SS Shop) coding guidelines?

StephenHonor: That's absolutely fine, do you want me to rename the other function above
also and references?

markguinn: I didn't even see that method. That's probably a holdover from darker days. I see that you've already corrected this, but if you have time to correct the other one too that'd be great. Also if you wouldn't mind squashing these into one commit I'd appreciate it.

markguinn:
Thanks, Stephen.

PR #389, opened 2015-11-19
Allow AddProductForm to be submitted via AJAX

adrian-stein:
To be used with PR on silverstripe-shop-ajax

### DIFF
--- a/code/product/AddProductForm.php
+++ b/code/product/AddProductForm.php
@@ -58,12 +58,12 @@ class AddProductForm extends Form {
 				$form->SessionMessage($cart->getMessage(),$cart->getMessageType());
 			}
 
-            $this->extend('updateAddToCart', $form, $buyable);
+			$this->extend('updateAddToCart', $form, $buyable);
+			
+			$request = $this->getRequest();
+			$this->extend('updateAddProductFormResponse', $request, $response, $buyable, $quantity, $form);
 
-            $request = $this->getRequest();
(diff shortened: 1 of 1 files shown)


markguinn:
Thank you @adrian-stein! One little thing - the code you've submitted uses spaces instead of tabs. It seems like a nitpicky thing but over time the codebase would get really messy if we didn't keep on top of things like that. Do you want to fix that yourself (you can just amend your commit and then force push) or would you prefer I did and merged it manually?

adrian-stein:
I can see if I can easily make the changes. Hopefully it is just the tab settings in my IDE!

markguinn:
Which IDE are you using? If you download an http://editorconfig.org/ plugin for it, you won't have to change the settings. It should just use tabs when you're in the shop folder and spaces everywhere else.

adrian-stein:
I am using PHP Storm. Im sure there is a setting somewhere in it. Just need to find it.

markguinn:
I believe it's under Code Style. PHP Storm definitely has an editorconfig plugin which can make the whole thing more painless.

adrian-stein:
I have installed the plug in. Do you have a .editorconfig file with all the code styles you would prefer to use so I make sure  have it set up exactly how you guys do it?

markguinn:
There should be one already in the shop folder. I just realized there probably isn't one yet in the shop-ajax module. Feel free to copy the one over from shop and include it in your PR.

adrian-stein:
Should be pushed now.

PR #395, opened 2015-11-25
Added hook for Ajax updateCartFormResponse

adrian-stein:


bummzack:
@adrian-stein Could you explain how you set up an AJAX enabled CartForm? For me, only the first update to the form works and subsequent updates won't work (eg. quantity values will never be applied), because fields are named differently. This is because the rendering of the AJAX response doesn't happen within the context of a `CartForm` and is thus missing the `CartEditField`.

I've got it working by changing the context to be the `CartForm` itself and by using `$forTemplate` as content in the `CartTable` template. But maybe you've got a setup that's better and works with the current implementation?

adrian-stein:
I have it with subsequent updates working. The only issue I am having which I am yet to fully address and find a solution is that on first load the table row uses CartEditField to create the the quanity and remove fields but after AJAX it renders with ShopQuantityField. Both work it just looks strange on update it adds extra features and changes things. 

The main change I made was to Cart.ss as below
```
<% require themedCSS(cart,shop) %>
<div data-ajax-region="CartTable" >
    <%include CartTable Editable=$Editable  %>
</div>
```
<code>CartTable</code> is the same I just extracted it into its own file as when it did the <code>pushRegion</code> method the whole cart was replaced inside itself. Passing Editable makes sure stays editable and I can keep the text boxes and remove checkboxes.

Due to the issue of the way the text boxes are rendered different on update I am open to a better solution that removes this issue. I just have not had the time to get around to finding one.

I am not qute sure what you mean when you say by "using <code>$forTemplate</code> as content in the CartTable template"?

bummzack:
@adrian-stein Thanks for the reply. I think the whole thing could be simplified quite a bit. In my opinion it would be optimal if the HTML output stays the same, whether it's an AJAX call or a normal HTTP request.

I even found a way that makes the `CartTable` template obsolete. My current solution looks like this:
```php
public function updateCartFormResponse(&$request, &$response, $form=null) {
    if ($request->isAjax()) {
        if (!$response) $response = $this->owner->getAjaxResponse();
        $this->setupRenderContexts($response);

        $data['Editable'] = true;
        $response->pushRegion('Cart', $this->owner, $data);

        if(self::config()->show_ajax_messages) {
            $response->triggerEvent('statusmessage', array(
                'content'   => $form->Message(),
                'type'      => $form->MessageType(),
            ));
            $form->clearMessage();
        }

        ShoppingCart::curr()->calculate();
    }
}
```
So instead of using `$this->owner->cart` as context (which is the current Order), I think it makes sense to use the `CartForm` instead. So just `$this->owner`. Then you can simply use `'Cart'` as the template which is already used by `CartForm` for rendering.

Seems to work really well…

bummzack:
Looks like I was too quick with my assumption. The `CartTable` is still needed. 

    $response->pushRegion('CartTable', $this->owner, $data);

And then in `CartTable` just put the following:

    $forTemplate

cart

adrian-stein:
I can get behind this idea. Although if we are now returning CartForm and not just the table we need to move the clear message before the pushRegion so the rendered form HTML does not contain the message, as below.

```php
	public function updateCartFormResponse(&$request, &$response, $form=null) {
		if ($request->isAjax()) {
			if (!$response) $response = $this->owner->getAjaxResponse();
			$this->setupRenderContexts($response);

			if(self::config()->show_ajax_messages) {
				$response->triggerEvent('statusmessage', array(
					'content'   => $form->Message(),
					'type'      => $form->MessageType(),
				));
				$this->owner->clearMessage();
			}

			$data['Editable'] = true;
			$response->pushRegion('CartTable', $this->owner, $data);

			ShoppingCart::curr()->calculate();
		}
	}
```

Also maybe we should add the cartTable.ss into the project so that people can use it as there is no documentation about how it works. Thoughts?

I am not sure about you but I have the following as the cart section on CartPage.ss. People will need to know where to add the <code>data-ajax-region="CartTable"</code> part as it is not obvious when I did it. I am not sure if we should include this or update the docs or request a PR somewhere. Ideas?

```html
<div data-ajax-region="CartTable">
    <% if Cart %>
        <% if CartForm %>
            $CartForm
        <% else %>
            <div>
                <% with Cart %><% include Cart Editable=true %><% end_with %>
            </div>
        <% end_if %>
    <% else %>
        <p class="message warning"><% _t('CartPage.ss.CARTEMPTY','Your cart is empty.') %></p>
    <% end_if %>
</div>
```

bummzack:
Sounds good. I think adding the `CartTable.ss` to the project is a good idea.

Regarding the status-message: You're right. I would even go as far and remove it even if `show_ajax_messages` is false. Eg.

```php
public function updateCartFormResponse(&$request, &$response, $form=null) {
    if ($request->isAjax()) {
        if (!$response) $response = $this->owner->getAjaxResponse();
        $this->setupRenderContexts($response);

        if(self::config()->show_ajax_messages) {
            $response->triggerEvent('statusmessage', array(
                'content'   => $form->Message(),
                'type'      => $form->MessageType(),
            ));
        }

        $form->clearMessage();
        $response->pushRegion('CartTable', $this->owner, array('Editable' => true));

        ShoppingCart::curr()->calculate();
    }
}
```
I think the cleanest way to add the `data-ajax-region` is to simply add it to the form in the `ShoppingCartAjax` Extension. Like so:

```php
public function updateCartForm(&$form, $cart) {
    $form->addExtraClass('ajax');
    $form->setAttribute('data-ajax-region','CartTable');
}
```
If we did it that way, there's not much to do for the user except installing the module to get an AJAX enabled cart.

bummzack:
There's another issue with the current `updateCartFormResponse` implementation. Recalculation of the cart should happen *before* pushing the region.

Here's my current version. I've also renamed `CartTable` to `CartFormAjax`:

```php
public function updateCartFormResponse(&$request, &$response, $form=null) {
    if ($request->isAjax()) {
        if (!$response) $response = $this->owner->getAjaxResponse();
        $this->setupRenderContexts($response);

        if(self::config()->show_ajax_messages) {
            $response->triggerEvent('statusmessage', array(
                'content'   => $form->Message(),
                'type'      => $form->MessageType(),
            ));
            $form->clearMessage();
        }

        ShoppingCart::curr()->calculate();

        $response->pushRegion('CartFormAjax', $this->owner, array('Editable' => true));
    }
}
```

markguinn:
This is great, guys. I'll trust your judgment on where to put this stuff. If it's strictly related to the ajax implementation lets keep it in shop-ajax though.

bummzack:
All the changes could go into shop-ajax. If Adrian agrees with the changes, I'd gladly create a PR for this.

adrian-stein:
It all sounds good except one problem. I have a quick add option on my cart page where people can enter a product number into a text box and add the product directly into the cart from the cart page it self. The issue with putting data-ajax-region onto the form is that if there is nothing in the cart the cart page does not render the form and thus not the attribute and then the cart with the added product does not have a region to put its html.

I was just testing your code with my setup so not sure if it is my set up, me mucking something up or the new code but im noticing now that the element with the attribute on is not getting replaced but instead having the html that is returned from the AJAX call being appending to it. https://github.com/markguinn/silverstripe-ajax/blob/master/javascript/base.js#L153 it appending so I get the complete rendered form with tags as a child of the original form tags which are not overwritten.

bummzack:
I guess the quick-add option isn't part of the form itself then? So it won't work with a form-submit (eg. Update cart) when JS is disabled? If that's the case, I'd treat it as a separate element all together and trigger updates of the cart with it?

In my setup it replaces the contents of the form with new content just fine (no duplicates or anything). I'll test with a vanilla install of shop too to make sure. I think the primary goal here should be to provide a drop-in module for an AJAX enabled shop. Users can then still dig deeper and customize/add functionality.

bummzack:
I just tested with a vanilla shop install, using:

"silverstripe/cms": "^3.2"
"burnbright/silverstripe-shop": "dev-develop"
"markguinn/silverstripe-shop-ajax": "dev-master"

AJAX updates of the cart work out of the box with the changes applied that were discussed in this thread (eg. added a separate template to render the form and updated the methods in `ShoppingCartAjax`).

I suggest we go ahead with this change and then try to figure out where Adrians problems arise?

adrian-stein:
Ok sounds good to me.

adrian-stein:
@bummzack Finally back on this cart thing. 

Just checking what happens with your project when you remove all the items from your cart. Does the Update cart button still appear?

bummzack:
@adrian-stein Yeah, the button is still there, with an empty cart message. When I refresh the page, the button is gone.

adrian-stein:
Cool. I was just making sure I was not going crazy. That still doesn't sit well with me so I might see if I can fix it if I get the time.

bummzack:
@adrian-stein It's not optimal, but I don't consider it a big flaw either. You get the "empty cart" message and when you try to enter checkout you'll get a "Your cart is empty" reminder as well.

There are several ways to fix this though. You could check the number of items in the cart and remove all unwanted form actions as soon the item-count reaches zero. 
Here's an example that simply removes all form actions (to be put into `updateCartFormResponse`):

```php
if($items = $this->owner->Fields()->fieldByName('Items')){
	if($items->getItemsList()->count() == 0){
		while($this->owner->Actions()->count() > 0){
			$this->owner->Actions()->pop();
		}
	}
}
```

This will still render the "Proceed to Checkout" link, since this is part of the outer template (`CartPage.ss`). Fixing all this would probably require quite a lot of template refactoring and is probably not a good idea for backwards compatibility… but it could be done in the 2.x branch.

Issue #405, opened 2016-01-06
Testing fails when symlinked because of absolute paths.


bummzack:
I'm often using symlinks for development. When I symlink the shop-folder, most of the unit-tests fail though.

I tracked the issue down to the following sources:
https://github.com/burnbright/silverstripe-shop/blob/develop/tests/ShopTest.php#L12
and
https://github.com/burnbright/silverstripe-shop/blob/develop/tests/cms/ProductBulkLoaderTest.php#L12

I'm wondering why `ShopTest.php` doesn't use something like: 
```php
include __DIR__ . DIRECTORY_SEPARATOR . 'test_config.php';
```
SilverStripe 3.1 has PHP 5.3+ as dependency, so `__DIR__` should be available.

The same could be applied to `ProductBulkLoaderTest`, or make use of the built-in `getCurrentAbsolutePath` method… 

Is this some legacy code (from pre SilverStripe 3.1), or is there a specific reason for the current implementation? The only thing I can imagine is something related to the CI Test runner, but I assume this should also work when changed to paths built using `__DIR__`?

bummzack:
Fixed with #406

PR #406, opened 2016-01-07
Fix for #405

bummzack:
Don't rely on silverstripe path, but rather use paths that are calculated relative to the file. Allows execution of tests when folders are symlinked.


bummzack:
Anybody have a clue what this Scrutinizer error means?

markguinn:
No but don't let Scrutinizer hold you back. The code coverage is notoriously fragile.

Issue #368, opened 2015-09-24
Issues with localization/translations.


bummzack:
There are some issues regarding localization of this module (I'm using 1.1.0).

 - [x] The translation files are mostly still in the old format (php files instead of yml). Yet using the legacy-adapter for the PHP files doesn't work (because `i18n::include_locale_file` doesn't exist). 

 - [ ] Existing YML files are incomplete (`en.yml`, `nn.yml`) or broken (`es_ES.yml` should be named `es.yml`, `nb.yml` throws a formatting error)

 - [x] Templates use the old syntax for translations (`<% _t('KEY', 'Default value') %>` instead of `<%t KEY 'Default value' %>`).

 - [x] Translations use sprintf (sequential) placeholder-replacement instead of the new variable-based replacement.

 - [x] Translation keys don't exist or are improperly linked. See **Example 1**.

 - [x] There are several duplicate translations. Instead of having `OrderInformation.ss.QUANTITY`, `OrderInformation_Editable.ss.QUANTITY`, `OrderInformation_PackingSlip.ss.QUANTITY` and `ProductCategoryItem.ss.QUANTITY` it would probably be sufficient to have `Order.QUANTITY` and use this identifier in several templates. There are also translations for `Order_statusEmail.ss'` and `Order_StatusEmail.ss'` in some of the lang-files, the first one is probably a typo (or legacy value) and can be deleted.

#### Example 1:

The template File [`Cart.ss`](https://github.com/burnbright/silverstripe-shop/blob/1.1.0/templates/Includes/Cart.ss) contains several translation that don't exist. Excerpt:

```html
<th scope="col"><% _t("PRODUCT","Product") %></th>
<th scope="col"><% _t("UNITPRICE","Unit Price") %></th>
<th scope="col"><% _t("QUANTITY", "Quantity") %></th>
<th scope="col"><% _t("TOTALPRICE","Total Price") %> ($Currency)</th>
```

There is no translation for `Cart.ss.PRODUCT`, nor for any of the other of these values. Instead the default value from the template is being used. One should be able to replace all the placeholder values with "Translation missing" and still never see that string!

----

I guess this is a known issue. What would be a good way to go about this? I guess starting with a proper language file for english (`en.yml`) should be the first step?

markguinn:
[...]

PR #408, opened 2016-01-08
Cleaning up translations in all PHP and template files.

bummzack:
A first stab at issue #368 

Updated translation syntax across all template and PHP files.
Tried to remove redundant and outdated translations… the resulting language file went from 770 lines to ~490 Lines.
Currently only `en.yml` has been updated.

Squashed commits:
 - Merged some more translations, re-collected all translations using the text-collector. 
 - Cleaning up translations.
 - Remove some more redundant translations.
 - Merged AccountNavigation and AccountPage.
 - Merged `Cart` and `ShoppingCart`
 - Created new en file using text-collector task.
 - Changed translation keys to camel case, some were renamed for clarity.
 - Implemented `i18nEntityProvider` interface for classes that need it.
 - Replace all `sprintf` based text-substitutions with the new placeholder syntax.
 - Use capitalized keys for order-status. Reasoning: It's a generated output, similar to `SINGULARNAME` etc.
 - Updated all templates to use new template syntax. Also merged several identical translations with the same key. Process still not complete though.
 - Update Order to allow translation of Order-Status.

### DIFF
--- a/code/account/AccountPage.php
+++ b/code/account/AccountPage.php
@@ -37,7 +37,7 @@ class AccountPage extends Page {
 		if($page = DataObject::get_one('AccountPage')) {
 			return $page;
 		}
-		user_error(_t('AccountPage.NO_PAGE', 'No AccountPage was found. Please create one in the CMS!'), E_USER_ERROR);
+		user_error(_t('AccountPage.NoPage', 'No AccountPage was found. Please create one in the CMS!'), E_USER_ERROR);
 	}
 
 	/**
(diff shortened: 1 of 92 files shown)


bummzack:
Any pointers why this test fails? It looks for `Default Addresses` in the output, and the string is clearly there.

Not sure if that assertion is really necessary though. Seems to be a fragile thing to check for a specific string like that, that could easily be changed via translation files.

bummzack:
@markguinn any objections against merging this for the moment? I'll make sure tests will pass and then I can base further translation work on this?

markguinn:
Go ahead, @bummzack that's fine as long as you circle back and fix that test at some point. Sorry I haven't had a chance to look into why it's failing yet.

PR #411, opened 2016-01-12
Add Locale information to Order. Fixes #403

### LINKED ISSUE #403: Order complete callbacks should be locale aware.
Maybe add a `Locale` field to `Order` to keep track of the order locale?
This is needed to ensure correct localization of e-mails and other content that might get generated upon order completion. It's also valuable for customer-relations.

Other things to consider are the Locale support of omnipay itself: https://github.com/thephpleague/omnipay/issues/279
But maybe these can be added/tracked individually.

The different translation modules (Translatable, Fluent) should be considered when implementing this.

bummzack:
Added locale db-field to order.
Added helper tools to get/set current locale.
Ensure locale is properly saved while an order is in `Cart` state.
Install the saved locale during order processing.
Install order-locale for ajax requests to ensure ajax response is rendered with the proper locale.

### DIFF
--- a/code/ShopTools.php
+++ b/code/ShopTools.php
@@ -13,4 +13,43 @@ class ShopTools{
 		return $field;
 	}
 
+    public static function get_current_locale()
+    {
+        if(class_exists('Translatable')){
+            return Translatable::get_current_locale();
+        }
+
+        if(class_exists('Fluent')){
+            return Fluent::current_locale();
+        }
+
+        return i18n::get_locale();
+    }
+
+    public static function install_locale($locale)
+    {
+        // If the locale isn't given, silently fail (there might be carts that still have locale set to null)
+        if(empty($locale)){
+            return;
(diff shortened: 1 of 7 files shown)


bummzack:
@markguinn This could probably also be merged into `develop`. What do you think?

markguinn:
I don't see any breaking changes there so it could go in the next minor version. Assuming you started this on the 2.0 branch, I'm not sure the best way to move it to develop. If we just merge it, we'll also merge all the history on develop-2.0 which we don't want to do. I think you could start a new branch and cherry-pick it though. I think this would be great to have on the main branch if you can figure out the git-acrobatics.

bummzack:
@markguinn ok I'll merge this one and try to cherry-pick it into `develop` as well if that's fine with you.

markguinn:
That's great. Thanks!

PR #417, opened 2016-01-28
Update account form extend

adrian-stein:


markguinn:
Shouldn't this still call the hook on the model until 2.0? Even if it gives deprecation warning the hook should be active until then.

PR #420, opened 2016-02-04
Update Order.ss for incorrectly closed tbody element

AntonyThorpe:
tbody needs to be open.
Thanks


bummzack:
There are other template issues that need to be fixed. I also noticed there's an invalid `colspan` in [`Order_Payments`](https://github.com/burnbright/silverstripe-shop/blob/develop/templates/order/Order_Payments.ss#L4). Would you be willing to check table consistency/markup for the other templates as well?

AntonyThorpe:
Yes will do tomorrow Cheers Antony

PR #421, opened 2016-02-04
Added extra extension points

adrian-stein:


bummzack:
I'd be interested to know about the use-case(s) for these? If it's for AJAX calls, we should follow a similar pattern as in the shopping-cart, where the proper response will be returned instead of redirecting to another page.

markguinn:
Me too. I'm with Roman that if we're going to put extension points here we should allow them to override the default response.

markguinn:
I'm going to manually merge this into `release-1.2.0` and add the response changes in the same pattern we used in ShoppingCart_Controller.

adrian-stein:
No, its not for AJAX. Its to send emails on updating of details and finalising orders. The points still work for what I need, I just wont send back a $response to continue to use the default response. The issue will arise if these points get built into the silverstripe-shop-ajax module then I would assume there would be issues as my code would be void...

bummzack:
@adrian-stein what makes you think it will invalidate your code? When you send an email in such a hook it should't matter if it was triggered via AJAX or regular HTTP request, right?

markguinn:
I have no plans of adding this to shop ajax personally. I think if someone did it should be in a separate extension so it could be removed if it broke something else you were doing without losing all ajax functionality. There's probably already too much going on in the one extension in that module as it is.

adrian-stein:
@bummzack, its mostly due to my understanding of extensions ( or lack of it I guess). Even though Mark is not planning on adding the hook to the other module I am not sure exactly what would happen if the code had 2 extension hook for the same point. Would it run both methods from the same hook? I assume it does not know which one to do first, therefore it may do the AJAX response return first and then miss mine altogether? or would it always do both.

markguinn:
My understanding is that it would run both no matter what.

adrian-stein:
Well that is fine then. I just was not sure how it would handle 2 methods one returning a response to then return AJAX and one not having a response so it should return the normal non AJAX way. I just am not sure which option it would take.

bummzack:
Mark is correct, the method from both extensions will run. You're passing the response as a parameter to the function, so if you're not assigning anything to response, your function will not have any impact on the AJAX response. 

Of course it can cause problems if there are several extensions that try to set a response though. In that case, the last one being executed will win.

PR #424, opened 2016-02-07
Revised Order Templates and associated CSS

AntonyThorpe:
Checked html in the Order Templates and revised where necessary.  Adjusted css where appropriate.  Removed div, paragraphs and headings as the order templates are also used in emails.


bummzack:
Looks good, thanks.

PR #425, opened 2016-02-11
Make example config overwrite shop/config.yml

NobrainerWeb:
Changed the After so changes in this file will actually overwrite any config settings in the shop/config.yml


markguinn:
Thanks, Thomas!

PR #429, opened 2016-02-18
Added toArray() onto map to prevent empty arrays

daykinandstorey:


markguinn:
Good catch. Thanks so much for your contributions!

PR #430, opened 2016-02-18
Fixed issue with address duplication due to blank or readonly fields

daykinandstorey:
The issue with this is that the readonly Country field which is added to the fieldlist (when there is only 1 selectable country) for an address causes the address to always duplicate.

This means you end up with a lot of addresses in your database even if you don't change your address.

The other issue with this is that blank fields cause the same issue as above because for some reason silverstripe doesn't populate the dataobject's record with blank fields. This means when it does the changed check it thinks things have changed even if they haven't.

### DIFF
--- a/code/checkout/components/AddressCheckoutComponent.php
+++ b/code/checkout/components/AddressCheckoutComponent.php
@@ -61,6 +61,14 @@ abstract class AddressCheckoutComponent extends CheckoutComponent
     public function setData(Order $order, array $data)
     {
         $address = $this->getAddress($order);
+        //if the value matches the current address then unset
+        //this is to fix issues with blank fields & the readonly Country field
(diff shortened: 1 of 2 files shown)


markguinn:
Travis seems to be running way behind today. I'm going to leave this a bit and see if it kicks in. Everything looks fine to me though.

bummzack:
@daykinandstorey Why did you close this one? If this fixes the duplicate address entries, then it would definitely be an improvement.

daykinandstorey:
I noticed an issue where it overrides the default details. I will be amending and redoing a pull request once I fix the issues.

bummzack:
Ah ok. Very good then ;) Cheers.

daykinandstorey:
I have redone the pull request with what I feel are better changes #431

PR #431, opened 2016-02-19
Fixed issue with address duplication due to blank or readonly fields

daykinandstorey:
This is the same as before, except I didn't really like the overriding of the core toMap functionality so removed that.

This meant I had to do the same thing but within the setData function so its now secluded to one place.

The additional bit that has changed is the getData section, it checks to see if an address is set first, if it is then it doesn't bother overriding it with the members default address (which is what it was doing). This is what resulted in blank fields keeping the default address.

### DIFF
--- a/code/checkout/components/AddressCheckoutComponent.php
+++ b/code/checkout/components/AddressCheckoutComponent.php
@@ -29,20 +29,32 @@ abstract class AddressCheckoutComponent extends CheckoutComponent
     public function getData(Order $order)
     {
         $data = $this->getAddress($order)->toMap();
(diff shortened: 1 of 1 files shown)


markguinn:
I like this solution better as well. Do you have time to write a unit test to cover this functionality (not the whole AddressCheckoutComponent, just the feature you added)? Tests of this part of the system are pretty thin and we're hoping to improve coverage. If not, that's ok. I'll still merge it once Travis has passed.

daykinandstorey:
Ahh sorry, I don't currently. I do agree with the lack of tests though.

PR #434, opened 2016-02-25
Changed some configuration (including defaults) around addresses in checkout

markguinn:
Currently the default is that new addresses are not associated with a Member's account. This seems wrong to me and is causing confusion from users. This PR corrects that default setting but also makes it easier to change the setting via YML config instead of having to override CheckoutComponentConfig with the Injector.


wilr:
Closing as ~2 yr old and needs a rebase. A lot of code in this has since changes (e.g can't see formfielddescriptions on AddressBook anymore).

PR #435, opened 2016-02-26
Names on shipping and billing address will now default to the main order names

markguinn:


code/model/Order.php
```diff
@@ -472,12 +467,30 @@ public function getBillingAddress()
...
+     */
+    protected function getAddress($type)
+    {
+        $address = $this->getComponent($type . 'Address');
+
+        if (!$address || !$address->exists() && $this->Member()) {
+            $address = $this->Member()->{"Default${type}Address"}();
+        }
+
+        if (empty($address->Surname) && empty($address->FirstName)) {
```
bummzack: Should this check/assign each field separately?

markguinn: I don't think so. If, for example they the shipping address had only the last name (which I've seen happen), we wouldn't want it to then inject the first name from the order because the customer probably did that for a reason. I think we only want it to inject if both fields are empty.

bummzack: Ok, fair enough.

PR #436, opened 2016-02-26
Added partial postcode support for postal codes

daykinandstorey:


markguinn:
Does this need to also check that the country is UK? I'm wondering if there might be other countries with similar post codes that might give false positives.

daykinandstorey:
I changed the code to check that the country is set to GB.

You are correct about this and I did think about doing it.

markguinn:
Would it be possible to clean up the commit history on this? Ideally we prefer to have each PR squashed into one descriptive commit, unless there's actually a good reason to have more.

daykinandstorey:
That's fine, the only reason its a bit mad is I merged it into my own repo. I will do that now

Issue #414, opened 2016-01-19
Use inline CSS styles for Emails


bummzack:
Something like [emogrifier](https://github.com/jjriv/emogrifier) should be used to inline order CSS instead of adding the styles in a `<style>` tag. Ensures better compatibility with Email clients (especially Outlook).

markguinn:
I've got a module for this: https://github.com/markguinn/silverstripe-email-helpers

Do you think it's worth integrating it out of the box? I suppose as long as do it in such a way that it can be disabled with the injector it makes sense.

bummzack:
Oh, I wasn't aware of this module.
I think styles should be inlined by default. In an ideal world, there would just be a CSS file that's getting inlined instead of the inclusion of styles via SilverStripe template.

Personally, I'd probably opt for using that module as a default for shop 2.0 and maybe clean up the templates a bit. Currently the Email template loads an external CSS *and* has styles in a `<style>` tag.

Does the module also allow inline images and other assets (base64 encoding)?

markguinn:
The module is really just a wrapper for Emogrifier. At the time I wrote it there wasn't support for that but Emogrifier may have evolved since then.

markguinn:
There's also this, which I haven't yet looked into: http://foundation.zurb.com/emails/email-templates.html

I don't think it's tied to the full Foundation framework. I wonder if we could base the new email templates on their templates and base CSS?

bummzack:
AFAIK emogrifier doesn't support inlining of images.
I think it's a good idea to look at these templates, or the mailchimp blueprints: https://github.com/mailchimp/Email-Blueprints

Maybe the ones you linked to could serve as a base for our own templates, but in the end it's just table-based layouts.

Regarding integrating your module into shop: If shop would use `Email::create` instead of `new Email`, it should be possible to just use `StyledHtmlEmail` with an Injector override, right?

markguinn:
That's what I was thinking, yeah.

bummzack:
What about the following procedure:
 - Where appropriate, update `new Class` with `Class::create` in the *develop* branch, so that users can already easily integrate your module.
 - At a later stage (2.0), rework Email templates and styles. Maybe even default to your module/emogrifier?

markguinn:
I think that's a great plan. And if we can find anything that does image inlining that'd be neat too.

markguinn:
[...]

PR #441, opened 2016-03-02
Updated OrderReceiptStyle.ss for Email CSS Resets

AntonyThorpe:
Uncommented the global resets in OrderReceiptStyle.ss and provided some base styles as recommended by seanpowell/Email-Boilerplate.   Also added a table#container wrapper to provide left/right padding in emails.  Thanks.

### DIFF
--- a/css/order.css
+++ b/css/order.css
@@ -3,6 +3,7 @@ h1.title {
     text-align: right;
     border-bottom: 1px solid #CDDDDD;
     text-transform: uppercase;
+    line-height: 1.5em;
 }
 
 .warningMessage {
--- a/templates/Includes/OrderReceiptStyle.ss
+++ b/templates/Includes/OrderReceiptStyle.ss
@@ -1,124 +1,50 @@
 <style>
-<!--
-/** Global resetting for Design **/
+	/** Global Resets for Email Design **/
+	/** Reference: https://github.com/seanpowell/Email-Boilerplate/blob/master/email.html **/
+	
 	html {
-		font-size:1em;
-		font-family:Verdana, Arial, sans-serif;
-	}
-	body {
-		font-size:12px;
-		padding:10px;
-		margin:0;
-	}
-	a img { border:0; }
-	
-	h1.title{
-		font-size:1.5em;
-		display:block;
-		text-align:right;
-		border-bottom:1px solid #CDDDDD;
-		text-transform:uppercase;
+		font-size: 1em;
+		font-family: Tahoma, Verdana, sans-serif;
 	}
 	
-	#Content {
-		text-align:left;
-		margin:auto;
+	body, table#container {
+		font-size: 12px;
+		line-height: 100% !important;
+		padding: 0;
+		margin: 0;
+		width: 100% !important;
+		height: 100%;
+		box-sizing: border-box;
+		-webkit-font-smoothing: antialiased;
+		-webkit-text-size-adjust: 100%;
+		-ms-text-size-adjust:100%;
 	}
+	
+	table td {border-collapse: collapse;}  /* Outlook 07 & 10 padding issue */
 
-	table#SenderTable{
-		width:100%;
-	}
-		table#SenderTable .sender,
-		table#SenderTable .meta{
-			width:50%;
-		}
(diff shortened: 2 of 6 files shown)


markguinn:
@bummzack do you have any thoughts on this? I know we'd talked some about reworking these emails based on MailChimp's templates or Foundation for Emails. Do you think this serves the same purpose and makes that task unnecessary? It looks to me like it would improve the quality of the default emails.

bummzack:
Yeah, this looks like a much better Email template. The problem is, that this PR seems to remove a lot of the inlined styles and relies solely on the inclusion of the external CSS? 
Or how did you add the order-styles to the Email @AntonyThorpe ?

Some of the points in #414 are still valid, although possible outside of the scope of this PR alone. I don't like having CSS in `.ss` files either.

Ideally we'd have inlined CSS from a CSS file, merged into the HTML via emogrifier. Could all of this be implemented in your Email helper module? Maybe even add inline-images as a feature?

markguinn:
True enough. It looks like PHPMailer (which my email-helpers module uses) can inline the images for you (http://stackoverflow.com/questions/1851728/how-to-embed-images-in-html-email) so one option would be to include this and add my module to the "suggested" list.

We could then require it in the 2.0 branch and swap out Email for StyledHtmlEmail in the order notifier. OR we could just integrate Emogrifier directly as it's a pretty simple affair.

bummzack:
I think we should require your module for 2.0. I think everybody wants to send properly formatted Emails that look good in most clients, unless they have a separate solution for Emails?

I'd keep things as they are handled currently for 1.x, so that your module is optional there. We can also discuss this matter with the other core-team members first.

markguinn:
Great. Did you see my note that I need to re-schedule the meeting for tonight?

bummzack:
Oh, no I didn't. Good to know. Where did you post that note?

markguinn:
On the Doodle poll. I've just sent you guys an email about it.

AntonyThorpe:
@bummzack the CSS in OrderReceiptStyle.ss is currently commented out so is not used.  You are right Silvershop currently relies on an external file, order.css (via order.ss), to provide all the email styling.  We could place the email resets in an external file as well.  Would this be a better approach?

Note: the comment element in OrderReceiptStyle.ss currently breaks the emogrifier process of the Silverstripe Email Helpers module.

markguinn:
I just created a PR against 2.0 that requires the email-helpers module. Now that I'm thinking about it, that could probably go in feature release as it doesn't really break anything. What do you guys think? Should I rebase that PR against master and it can go into 1.3 along with this?

bummzack:
@markguinn Sounds good to me.

markguinn:
Ok. I'm going to go ahead and merge this and rebase my work on top of this.

markguinn:
@AntonyThorpe - I may have merged this prematurely. The default emails now look a lot less nice than they used too.

https://www.dropbox.com/s/7tcfefmnd7j3n70/Screenshot%202016-03-23%2015.02.04.png?dl=0

vs.

https://www.dropbox.com/s/29gqxookc1qhwm1/Screenshot%202016-03-23%2015.03.46.png?dl=0

What do you think about adding back some of the old styling on top of these resets? (I'm happy to do the work, just curious if there was more to your master plan). Is this about how your emails are looking?

AntonyThorpe:
@markguinn Sorry, that is quite different.  My emails are coming through like your second link, fully styled.  The plan is to get the EmogrifiedSmtpMailer class of Silverstripe Email Helpers working without the comment element in OrderReceiptStyle.ss causing errors.  Then I thought a left/right margin was needed and some base email resets.  Will take a closer look and get back to you.

markguinn:
I had a few merge conflicts in pulling this into 2.0 and merging it with my emogrifier work. I might have bastardized your changes a bit.

AntonyThorpe:
@markguinn Have worked it out.  I thought that the commented out styles in OrderReceiptStyle.ss were not used in emails, but they are.  Also, order.css is not used for emails (my comments above are incorrect). Will send through a pull request to add back the reset styles.

PR #446, opened 2016-03-10
Validate shop account form

adrian-stein:


bummzack:
It seems like you somehow merged validation from `ShopAccountFormValidator` directly into the submit function. Having a proper validator would be more useful in my opinion though.
I've fixed Member_Validator in the silverstripe core (https://github.com/silverstripe/silverstripe-framework/pull/5098), so all this will probably obsolete once it lands (most likely in 3.4).

adrian-stein:
I dont think the validator even being used. If it is there it should really be used.

bummzack:
Oh, just minor improvements. Updating the query (`DataObject::get_one('Member', "$field = '$uid' AND ID != " . $currentmember->ID)`) to more recent syntax and not use `sprintf` for translation output. Although the latter is already changed in the 2.0 branch. I think in terms of functionality, the validator does what it should. Maybe a hook for extensions could be added, but not really mandatory since this should all be solved via silverstripe framework soon.

adrian-stein:
Sounds good.

bummzack:
Fixed via #459

PR #450, opened 2016-03-16
Pass more data through to the payment processor if needed

markguinn:
Does anyone see a problem with this? It's helpful for the Braintree integration I'm working on because it means the payment gateway gets to see fields in the input that aren't prefixed with `OnsitePaymentCheckoutComponent_` - in this case `payment_method_nonce`.

The worry would be this gives a bit more surface for an attacker to say inject things like the amount (although that specific one is covered). The PurchaseService does explicitly set some things and I don't see any obvious attack vectors.

### DIFF
--- a/code/checkout/PaymentForm.php
+++ b/code/checkout/PaymentForm.php
@@ -107,7 +107,7 @@ class PaymentForm extends CheckoutForm
         if ($components->first() instanceof CheckoutComponent_Namespaced) {
             foreach ($components as $component) {
                 if ($component->Proxy() instanceof OnsitePaymentCheckoutComponent) {
-                    $data = $component->unnamespaceData($data);
+                    $data = array_merge($data, $component->unnamespaceData($data));
                 }
             }
         }


bummzack:
Hm. I don't really understand why the data wasn't merged in the first place? It seems odd that the data is collected and then discarded if the component is of type `CheckoutComponent_Namespaced`? Personally I've never used this checkout-component, but I see no harm in your PR… I'd even consider an improvement.

And it doesn't change the "injection" vulnerability, does it? A malicious user could just as well create extra fields with the namespace, which would result in the same situation. It's important that shop merges in the order-information last (thus overriding any malicious input the user might have entered). The omnipay gateway is then responsible for culling all the data that isn't needed for the gateway.

markguinn:
You're absolutely right of course. I was thinking then "unnamespacing" there was actually validating against a set of real fields but it's not. It just removes the namespace. So any vulnerability would be present either way. Thanks for the review, Roman.

PR #452, opened 2016-03-16
Improvements to the addressbook UI

markguinn:
This allows you to create an interface like this:

https://www.dropbox.com/s/3ff0dlymiopdwps/Screenshot%202016-03-16%2016.02.21.png?dl=0
https://www.dropbox.com/s/3pbjzl5p339hn92/Screenshot%202016-03-16%2016.15.15.png?dl=0

I think that's better than the dropdowns you got by default and this is something we've done on all the sites we've built anyway. Any objections to just adding this as the default UI?

### DIFF
--- a/code/account/AccountPage.php
+++ b/code/account/AccountPage.php
@@ -19,6 +19,8 @@ class AccountPage extends Page
      * Returns the link or the URLSegment to the account page on this site
      *
      * @param boolean $urlSegment Return the URLSegment only
+     *
+     * @return mixed
      */
     public static function find_link($urlSegment = false)
     {
@@ -31,6 +33,8 @@ class AccountPage extends Page
      *
      * @param int|string $orderID    ID of the order
      * @param boolean    $urlSegment Return the URLSegment only
+     *
+     * @return string
      */
     public static function get_order_link($orderID, $urlSegment = false)
     {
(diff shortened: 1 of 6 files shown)


bummzack:
Shouldn't this go into 2.0?

markguinn:
It can. I don't think it has to, but that might make more sense.

bummzack:
Hm yeah it doesn't change public API much. I don't understand what the "Make default XY" Buttons are doing though. How do you assign another address as the default? Remove via Trashcan-Icon, then assign somehow? Sorry, but I don't get it from the two screenshots ;)

markguinn:
The trashcan icon deletes the address from your addressbook (but not from previous orders). The "Make default XY" button sets whichever address you click it for to be the default shipping or billing address (meaning the one that is prepopulated in checkout). Does that make sense?

bummzack:
Oh, I see. So every Address is listed as a "card" and can have the "Default Shipping Address" and/or "Default Billing Address" label? So an Address could theoretically have both labels, right?

> deletes the address from your addressbook (but not from previous orders).

So, it doesn't delete the Address, but just *unlinks* it, right?

markguinn:
Right on both fronts.

bummzack:
Seems like a good change… although a bit more bulky than previously. But I guess nobody is going to have dozens of Addresses. And the possibility to "delete" an Address is a nice addition.

markguinn:
It's easy enough to change back to the old UI too. The form with the two dropdowns is still present in AccountPage_Controller so you could just override that template and use $DefaultAddressForm. I think I will make this a 2.0 change though instead of 1.x. It could easily disrupt the UI of an existing site.

bummzack:
Sounds good.

PR #461, opened 2016-03-18
Clear the cart/session for every test-suite that uses the shopping-cart.

bummzack:


bummzack:
@markguinn I added an initialization step that clears the cart/session for every test that makes use of `ShoppingCart`. This should ensure that tests operate with a clean cart and don't get data from another test. What do you think?

PR #463, opened 2016-03-22
Remove deprecated methods.

bummzack:
Removed ShopBuyableException.
Cleaned up `canPurchase` implementations in the process (no longer throw/catch exceptions there).
Use config API for WeightShippingModifier. Removed static setter.
Fixed tests to no longer use deprecated methods.


bummzack:
@markguinn What do you think? We should be fine to remove the parts that were marked as deprecated for 2.0, right?

markguinn:
Yeah I think that's great. I also love to see those docblocks getting healthier. Thanks!

PR #465, opened 2016-03-23
Requires email-helpers. Uses StyleHtmlEmail (Emogrifier) by default

markguinn:


markguinn:
Interesting. This may need a bit more work and thought. Apparently Emogrifier requires PHP 5.4. That means our tests fail against PHP 5.3. Which makes me think this actually should go in 2.0 instead of 1.3. AND that 2.0 should now require PHP 5.4 as well. Any objections to raising the minimum php version in 2.0 @bummzack @wilr or @anselmdk?

bummzack:
Ah, that's too bad. I agree to put this into 2.0 only and raise the minimum PHP version accordingly. We'll be slightly off the requirements imposed by SilverStripe itself, but only until 4.0 lands. Then the minimum version will be 5.5 AFAIK.

markguinn:
It's a little tempting to bump to 5.5 as well but I imagine we'll have to do a major release to support 4.0 as well so we can do another bump then to match them. No reason to cause anyone extra trouble before we actually have to :)

bummzack:
Yeah, I totally agree.

anselmdk:
I agree, let's bump to `5.4` for now.

PR #468, opened 2016-03-23
No longer use deprecated methods (snake_case) of the omnipay module

bummzack:
See https://github.com/burnbright/silverstripe-omnipay/pull/104.
Also fix deprecated `Member::get_unique_identifier_field()` usage.


markguinn:
Packagist still hasn't updated. I've asked @jedateach for access to be able to check on and hopefully fix that. This is on hold until that's fixed.

bummzack:
Yeah. Good thing 2.0 still uses the 1.x branch of omnipay, otherwise all PRs against 2.0 would fail CI.

PR #474, opened 2016-03-31
Fixed a null pointer issue

fonsekaean:
Is there is an attribute type with no options then it doesnt return a dropdown. and that breaks the Product with with a 500 error


markguinn:
CI just failed because of a github issue I think. Will merge as soon as everything passes.

fonsekaean:
thanks @markguinn

PR #477, opened 2016-04-11
Added French translation, pulled from Transifex.

bummzack:


bummzack:
@markguinn Seems like there's an issue with Scrutinizer? Is it because emogrifier isn't referenced with a proper version in the composer file?

Issue #427, opened 2016-02-18
SS 3.3 RC2 breaks variations


bbowden:
Using SS-Shop 1.2 with SS 3.3 RC2 and I can choose an attribute but trying to Add Variation gives an error:
Error at line 55 of /Users/macbookbruce/websites/felters/framework/model/connect/DBConnector.php
Reverting back to SS 3.2 and the error goes away.

cheers
bruce

markguinn:
Thanks for reporting this. I've seen similar issues with SS 3.3 and I'm not sure what's causing that. If you find out let me know! If not, know that it's on my radar so it'll hopefully be fixed shortly.

markguinn:
@bbowden are you using https://github.com/silverstripe-labs/silverstripe-fulltextsearch on this site? If so you should update to a newer version (~2.1.1). If not we can debug from there.

bbowden:
Hi Mark,
I’m not using that module.  The site is unusual in that SS-Shop has become a booking tool rather than a selling tool but all changes are done outside the shop module code. My development site is on my Mac so, with luck, that may make the debugging a little easier.

cheers
Bruce

markguinn:
Are you using any other module that might wrap the database connector? For example, pre-1.0 versions of my clockwork module?

bummzack:
I've just encountered this issue as well. It seems to look for a `ProductVariation_Live` table, but that doesn't exist.
I suspect this is related to the changes to `Versioned` that were [introduced with 3.3](https://docs.silverstripe.org/en/3.3/changelogs/3.3.0/). 

To me it looks like this declaration: `Versioned('Live')` didn't previously require a `_Live` table, since the only version was `Live`, but with 3.3 this seems to be the case (although dev/build doesn't create them, so this might be a framework bug).

Using the `Versioned` extension for just one "Stage" seems questionable to me anyways… why is `ProductVariation` versioned with just `Live`? Instead of `Stage` and `Live`?

bummzack:
This is definitely a framework issue. I'll open an issue over there and reference it back here.

markguinn:
I think the idea was that we wouldn't need to deal with publishing changes, but we'd still have a record so, for example, historical orders would display the correct data from when the item was purchased. It's been nothing but trouble for me though.

markguinn:
Once 4.0 drops it should be much easier to just make this versioned with Stage/Live I think.

bummzack:
[...]

PR #478, opened 2016-04-14
Added stage and live both to the versioning

fonsekaean:
without these configs it doesnt create a ProductVariation_Live table and results in errors.

"[User Error] Uncaught SS_DatabaseException: Couldn't run query: SELECT "Version" FROM "ProductVariation_Live" WHERE "ID" = ? Table 'getmagicbullet.productvariation_live' doesn't exist"


bummzack:
That's a bug in the framework that's causing this. Framework should be fixed, not `ProductVariation`. See #427

PR #479, opened 2016-04-14
Enforce order recalculation before marking as Paid.

bummzack:
Ensure order is recalculated before placing the order or before completing a payment.

### DIFF
--- a/code/checkout/OrderProcessor.php
+++ b/code/checkout/OrderProcessor.php
@@ -167,6 +167,9 @@ class OrderProcessor
     public function completePayment()
     {
         if (!$this->order->Paid) {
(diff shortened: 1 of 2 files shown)


bummzack:
Unit-tests fail now, since master also uses silverstripe-omnipay 2.0.x-dev. This is probably because the dev-dependencies do not allow silverstripe-omnipay 1.x at all, because they require omnipay 2.x gateways…

The tests fail because of deprecation warnings. Should we disable these in PHPUnit? It isn't good practice to do that though?

Or can we change master to use silverstripe-omnipay 1.x instead? With the current roadmap, only shop 2.x will be compatible with omnipay 2.x

@markguinn thoughts?

PR #485, opened 2016-04-21
Make shop 2.0 compatible to Omnipay 2.0

bummzack:


bummzack:
@markguinn Is there a stable version for pelago/emogrifier that could be used in the email-helper dependencies? It seems this is the reason for the scrutinizer errors. Maybe you could also have a quick look at this PR to check if everything is fine from your perspective?

markguinn:
Last time I checked I don't think there was OR there was something in master that was needed for it all to work. Maybe that's not the case now?

bummzack:
I just checked the history and it seems like you added the composer dependency on July 31, 2014. Emogrifier `1.0.0` was released Oct 15, 2015, so changes are good that it would work with `~1.0` as version constraint?

markguinn:
Just updated the constraint in master. Assuming tests pass I'll tag a new version of email-helpers.

markguinn:
Ok, 1.2.1 is out. Might take packagist a little while to pick it up though.

bummzack:
Thanks Mark. Seems like everything is fine now. Should I merge this?

markguinn:
Go for it!

PR #490, opened 2016-04-28
Added ability to modify product Image method by extension.

hpeide:


markguinn:
What's the use-case for this? (not saying it's not necessary, just wanting to understand it)

hpeide:
In my case the customer want to show more than one image per product. So I had to create a many_many relation for images. To make the first image appear in the cart, checkout etc. I had to override this method and return Product->Images->First() from the updateImage. It may be bad practise?

bummzack:
I think this should be solvable via templates? Eg. by using `$Images.First` or `$Images.Sort('SortOrder').First` in your templates. Or have you tried that already?

hpeide:
Yes, agree, but somehow I can't get it to work and i don't figure why. Created a gist to show how I done things:
https://gist.github.com/hpeide/ef62e676a9a7dcef76f106f684c2bda2

There is also numerus templates to edit to adjust the thumbnail, is it a idea to seperate the thumb into an own template? Eg. CartThumbnail for Cart and SideCart, and Order_Thumbnail for OrderAdmin_Content_ItemLine, Order_Content_ItemLine and Order_ItemLine.

bummzack:
I agree that it's cumbersome to create custom templates to just modify the Image. 
Not sure if stripping them out into include templates is a good solution… (if we go further down the road with this approach we end up with separate templates for pretty much every HTML tag). But maybe we could identify some common blocks that would be worthwhile as a separate template?

I'm going to merge this into master for now until there's a better solution. 
@hpeide it would be great if you could look into the templates… or at least add some information to the extension hooks readme ([docs/en/02_Customisation/Hooks.md](https://github.com/silvershop/silvershop-core/blob/master/docs/en/02_Customisation/Hooks.md))

PR #494, opened 2016-05-05
Make ShoppingCart use config and Injector

riddler7:
Left everything else untouched so there will be no backwards breaks.


markguinn:
I think this makes a ton of sense. Thank you!

PR #495, opened 2016-05-09
Added logic to prevent upscaling product images

markguinn:


markguinn:
@bummzack - could you give this a quick review? Do you see any problems with doing this?

code/product/ProductImage.php
```diff
@@ -7,39 +7,73 @@
...
+     *
+     * @return Image
+     */
+    public function getImageAt($width = null, $height = null, $upscale = false)
     {
+        if (!$this->owner->exists()) {
+            return $this->owner;
+        }
+
+        $dim = explode('x', $this->owner->getDimensions());
```
bummzack: Why not use `$this->owner->getWidth()` and `getHeight()`?

code/product/ProductImage.php
```diff
@@ -7,39 +7,73 @@
...
+
         if ($width && $height) {
-            return $this->owner->SetSize($width, $height);
+            return $dim[0] < $width && $dim[1] < $height && !$upscale
+                ? $this->owner
+                : $this->owner->SetSize($width, $height);
```
bummzack: We should use the new image API, since the methods used here are deprecated for 4.0.

- `SetSize` => `Pad`
- `SetWidth` => `ScaleWidth`
- `SetHeight` => `ScaleHeight`

code/product/ProductImage.php
```diff
@@ -7,39 +7,73 @@
...
         } else {
             if ($width) {
-                return $this->owner->SetWidth($width);
+                return $dim[0] < $width && !$upscale
+                    ? $this->owner
+                    : $this->owner->SetWidth($width);
             } else {
-                return $this->owner->SetHeight($height);
+                return $dim[1] < $width && !$upscale
```
bummzack: Since you're using `$height` for scaling, this should probably read: 

    return $dim[1] < $height && !$upscale

code/product/ProductImage.php
markguinn: The main reason is to not have it run the calculations twice. Last time I checked the dimensions were not cached in any way.

code/product/ProductImage.php
markguinn: Good call.

code/product/ProductImage.php
bummzack: Is the string concatenation/explode via `getDimensions` really faster than two individual calls for `getHeight` and `getWidth`?

I'd rather have better readable and more stable code (what if the joining string changes from `x` to `×` or something else?). Using `getWidth` and `getHeight` is also in line with what is being used in the framework code (look at the usage of `getWidth` and `getHeight` vs. `getDimensions`)

markguinn: Makes sense to me.

PR #512, opened 2016-06-29
Improve the way variation options are generated

pmhou:
Page load is exponentially slowed on products with multiple attribute types due to the hidden input field that's generated by Variation Form. This fix has reduced the script execution time up to this point by over 50% on my local tests and reduces further as the number of product variations increases.


bummzack:
Awesome, thanks.

PR #519, opened 2016-07-05
Optional Shipping

wildflower:
I've been working with subscription products and so don't need the
shipping address form and the validation that comes with it.

I've added this switch in the Checkout Config to turn the shipping
address off:
CheckoutConfig:
ShippingOptional: true


bummzack:
I'm hesitant to add this… you can simply create your own Subclass of `CheckoutComponentConfig` with your desired configuration and then enable this one instead. So in your config file you would then add:

```yml
Injector:
  CheckoutComponentConfig:
    class: MyCustomCheckoutComponentConfig
```

Otherwise we'd have to add config flags for every possible checkout-component in the long run.

wildflower:
thanks, I thought there was probably a better way architecturally.

I haven't tried using Injectors before

PR #521, opened 2016-07-14
Correct ProductImage to be scaled by Height

beeonlinenz:
Was scaling the image width with the height setting


bummzack:
Good catch, thanks!

PR #522, opened 2016-07-21
Updated 02_Upgrading.md

AntonyThorpe:
Added a few notes to the Upgrading instructions following upgrading a site to v2.  Thanks.


bummzack:
Great, thanks!
We could also include a link to the Omnipay 2.0 changelog: https://github.com/silverstripe/silverstripe-omnipay/blob/master/docs/changelogs/2.0.md

AntonyThorpe:
Thanks, will do.

PR #524, opened 2016-07-21
Updated example_config.yml

AntonyThorpe:
Added an example of overriding Product's $db array with an adjusted data type.  Thank you.


bummzack:
Thanks!

PR #529, opened 2016-08-01
fontawesome is not using the icon class anymore

oetiker:
see also https://github.com/FortAwesome/Font-Awesome/wiki/Upgrading-from-3.2.1-to-4

### DIFF
--- a/templates/Includes/AccountNavigation.ss
+++ b/templates/Includes/AccountNavigation.ss
@@ -8,25 +8,25 @@
 
             <li>
                 <a href="{$Link}">
-                    <i class="icon icon-list"></i><%t AccountPage.PastOrders 'Past Orders' %>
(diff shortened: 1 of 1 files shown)


bummzack:
I agree to add the Font-Awesome classes, but I suggest to keep the old ones as well. The first reason is to not break someones existing template and the second reason is that there are other libraries that use `.icon`. An example for this is [fontello](http://fontello.com/) which I use for most projects (because I can have a subset of Font-Awesome or other icon-fonts **and** I can add my own custom symbols).

oetiker:
you mean keeping the old fontawesome names on top of the current ones ?

bummzack:
Yes, instead of replacing the existing classes, keep them. Eg. `icon icon-list fa fa-list`

PR #534, opened 2016-08-04
New feature: Order Status Email Notifications

AntonyThorpe:
This pull request enables an opt in to order status email notifications.  Upon a change of order status, an alert email is automatically sent to the customer (also works for guest users).  Few questions/points listed below for your consideration please:
- This pull request closely follows the design of the existing code by creating an OrderStatusLog, which triggers the sending of an email via the OrderEmailNotifier class.  This is out of step with other email notifications in Silvershop that are simply sent directly from the OrderEmailNotifier class.  Would skipping the logging step be a better approach?
- Have focused purely on creating the email notification feature and avoided addressing the purpose of the OrderLogStatus class (see Silvershop/core#306 for a discussion).     
- Have used the settings in `en.yml` to populate the body of the email.  Would it be better to create a new Model and allow a user to add the note from the CMS?

Thanks


bummzack:
Looking great, thanks!
I like the logging step. I even think it's mandatory to have as a fallback/reference. Maybe a reference could be added to both, email and log-entry, so that the log-entry can be found quickly when somebody mentions the reference from the update? Maybe status and order-reference are enough though… 

Having the status-text in the locale file seems fine as well. These will remain rather short, I assume? I guess a text-field used for all status-mails could be added to the CMS to inject some content (like contact-information). But this is potentially another change that would be useful for all Emails sent by shop?

Pinging @wilr, @markguinn, @anselmdk to also have a look.

markguinn:
This looks excellent to me.

wilr:
👍  from me. Tests pass. @bummzack You want to merge if your feedback is actioned.

bummzack:
Thanks Antony. Further improvements can be made with more PRs, if necessary.

AntonyThorpe:
Thank you @bummzack.  Will follow up with a pull request to show the Order Status Logs alongside the Orders in the CMS.

bummzack:
@AntonyThorpe Hey Antony, thanks again for your work on this. I've finally had some time to look into your changes in detail.

I'm planning to release 2.1 shortly, including the status-log feature. There's one thing I'd like to address first though: Currently, there's a potential that a status-change can be sent multiple times, either by writing the `OrderStatusLog` object or the `Order` object. It's currently not possible in the CMS, but I could imagine that an `Order` could have other editable fields than just `Status` and thus would trigger the `onAfterWrite`.

My suggestion is the following: Check in the `onBeforeWrite` hook whether or not the status has changed and set a flag for the change. Then only create the log entry when the status has actually changed.

As for safeguarding a single `OrderStatusLog` entry, I'd probably initialize `SentToCustomer` to `false` and only send the Email if it's false. When sent, set it to `true`. Currently it's just being initialized with `true`, but it's not certain that the email was actually sent?

If it's ok with you I'll create a PR with my suggested changes, so that you can have a look?

bummzack:
Proposed change is here: #545
Thoughts?

AntonyThorpe:
@bummzack Thank you for contacting me.  Am I following you correctly here? Currently in the master branch, if say a Silvershop module, accesses `onAfterWrite` of the Order class and re-writes the same order (saved twice) then two OrderStatusLogs are created and two email notifications are sent.  These changes to the Order class fix this problem, as an `OrderStatusLog` will only be created when the status of an order has actually changed from what it was previously.  

With the `OrderStatusLog` class, yes, `SentToCustomer` needs to default to `false`.  I had interpreted it as indicating that a notification needs to be sent.  The `flagSendEmail` now takes on this role and the `SentToCustomer` prevents additional notifications been sent in the future. 
Therefore, can I please suggest:
```php
// Order class in onAfterWrite
$log->flagSendEmail = true;  // $log->SentToCustomer defaults to false

// OrderStatusLog in onAfterWrite
if ($this->flagSendEmail && !$this->SentToCustomer) {
    OrderEmailNotifier::create($this->Order())->sendStatusChange($this->Title, $this->Note);
    $this->SentToCustomer = true;
}
```
What do you think?

Also, is there a need to write the `OrderStatusLog` to save `SentToCustomer` property within the `onAfterWrite` function?

bummzack:
@AntonyThorpe Yes, that's exactly what my proposed change is about. Prevent triggering multiple log entries/emails.

I'm setting `SentToCustomer` in the `onBeforeWrite` hook, so that it gets written to the DB once writing is done. I think your change to `if ($this->flagSendEmail && !$this->SentToCustomer)` doesn't work, because once `onAfterWrite` is reached, `SentToCustomer` should already be set to true (in onBeforeWrite)

I generally dislike triggering a new write in the `onAfterWrite` hook, as this could easily lead to an endless loop.

As of now it's implemented like this:

1.  `onBeforeWrite`: Check if `SentToCustomer` is `false`, if that's the case, flag for sending an email and set `SentToCustomer` to `true`.
2. `onAfterWrite`: Check if `flagSendEmail` is set, if yes send the email.  Since `SentToCustomer` was set to `true` in the `onBeforeWrite` hook, is is now persisted as `true`.

I think that approach is pretty stable, because if something fails during writing of the DataObject, the `onAfterWrite` hook won't be reached and no Email will be sent. Also the `SentToCustomer` param should not have been persisted… the additional flag is kinda ugly though.

The other option would be to ditch the flag and handle everything in `onAfterWrite`, but that would lead to the case where `onAfterWrite` triggers another write.

AntonyThorpe:
@bummzack thanks for clarifying the use of the onAfterWrite function.  Hey, do you think this is the right place to create an OrderStatusLog from within the Order class, and then an email notification within OrderStatusLog?  Would onBeforeWrite be more appropriate (in both cases) as there is a need to set properties?

bummzack:
I think  `onAfterWrite` is fine, because at that point we can be certain that the order/log-entry have been persisted to the DB. I don't mind the flags too much, as long as they aren't everywhere ;)

AntonyThorpe:
Another question for you please @bummzack.  Should the sending of a email notification be independent to the creation of a OrderStatusLog?  Within the onAfterWrite function of the Order class the OrderStatusLog could be created after sending an email notification.  Better pattern?

bummzack:
You're right, that's actually a better pattern. As long as we're only creating log-entries from within Order that is. But since it's named **Order**StatusLog, that's probably a safe thing to assume. 

I'll update the PR accordingly, then?

AntonyThorpe:
Yes I think this is a good idea.  I can give it a crack for you if you want over the next week?

bummzack:
I've updated the PR #545

AntonyThorpe:
@bummzack have tested the improve-order-state-log branch with both a Guest and Member and it works as expected.  However, have not tested the scenario where the order is written more than once.  Thanks.

bummzack:
@AntonyThorpe Yeah, it's something that can't currently be triggered manually in the CMS. But I've added unit-tests that simulate this. I'll merge the PR then. If you find further improvements, feel free to PR again.

PR #535, opened 2016-08-10
Visibility of Order Status Logs within the CMS 

AntonyThorpe:
Provides visibility of the Order Status Logs in a tab on the orders folder.  Thank you.


bummzack:
Awesome. Thanks a lot.

PR #536, opened 2016-08-11
Updated example_config.yml for log_status settings

AntonyThorpe:
Added log_status settings example for order status email notifications


bummzack:
Thanks!

PR #537, opened 2016-08-11
Update forms extensions

a2nt:


bummzack:
There's already an extension-point for `ShopAccountForm`, it's `updateShopAccountForm`, called from within the `ShopAccountForm` constructor.

Regarding your extension-point to modify the default-address-form: Can you give me a use-case for this? Are you adding other addresses than Shipping and Billing to your Order?

a2nt:
I'd like to make it bootstrap compatible so I will need to set custom form templates

bummzack:
Wouldn't that be a global change to `Form`? Or do you have different templates for every form?

a2nt:
Nope, I have BootstrapForm class for frontend which I use to setup custom templates on frontend side. Global change of the Form class will make changes in the CMS.

bummzack:
Can't you just put a custom `templates/Includes/Form.ss` inside your theme for custom Form markup? As long as it's part of your theme, it won't be used within the CMS…

a2nt:
There's also extra functions like appending fields, placeholders and etc

bummzack:
I see, please remove the `updateEditAccountForm` hook, as it is redundant. Then this is good to merge.

a2nt:
I will add updateForm hooks directly to Form objects and will send you a request, thank you.

bummzack:
I think we should decide on the path of action first by coming to a conclusion in #538

PR #541, opened 2016-08-23
colgroup should be col

oetiker:
the structure for setting properties on table columns is `<colgroup><col class="a"><col class="b"></colgroup> ... `. Column Groups without cols make little sense.


bummzack:
Thanks!

PR #543, opened 2016-09-05
Pass quantity to findOrMakeItem()

uniun:
Pass quantity to findOrMakeItem() so it can be passed to Buyable::canPurchase()


bummzack:
Seems like a good change, but wouldn't you also have to pass the `$quantity` to `createItem`? Otherwise you'll pass a quantity, but just create one item?

bummzack:
Thanks. I'll fix the `createItem` thing shortly.

PR #545, opened 2016-09-13
Only create an OrderStatusLog entry when the status has changed.

bummzack:
Ensure status-log email gets sent only once.
Added unit-tests.


beeonlinenz:
I have a website where customers and admin are getting two email confirmations with a paid credit card order (DPS) - am I correct in thinking this resolves that issue?

bummzack:
@beeonlinenz No, this isn't related to confirmations. It's only affecting the new Order Status Log (which is only available in the master branch as of now)

PR #547, opened 2016-09-20
Updated composer.json

AntonyThorpe:
Updated markguinn/silverstripe-email-helpers to v1.3 (latest version).  Thank you.


bummzack:
Just FYI: This wasn't really necessary, since the version constraint `~1.2` will match all `1.x` versions (but not `2.x`). Installing with `~1.2` will also install `1.3`, `1.4` etc.

Issue #443, opened 2016-03-05
canPurchase decorator methods not working


fonsekaean:
The canPurchaser decorator methods doesnt work on the Product and ProductVariation, as it is calling extendedCan and the canPurchaser decorator function doesnt get the quantity. 

Updating the code in Product->canPurchase method from 

```
// Standard mechanism for accepting permission changes from decorators
        $extended = $this->extendedCan('canPurchase', $member, $quantity);
        if ($allowpurchase && $extended !== null) {
            $allowpurchase = $extended;
        }
```

to 

```
// Standard mechanism for accepting permission changes from decorators
        $extended = $this->extend('canPurchase', $member, $quantity);
		if ($allowpurchase && $extended !== null) {
            $allowpurchase = $extended[0];
        }
```

And doing the same on product variation solves the issue. 

Thanks

bummzack:
I've used `canPurchase` in extensions before and it works. I guess the main issue here is, that the `$quantity` parameter is being discarded? So can-purchase checks that involve quantity will fail…

Since `extend`, as well as `extendedCan` filter out `null` values, I guess the code could be rewritten as follows:

```php
$permissions = $this->extend('canPurchase', $member, $quantity);
$permissions[] = $allowpurchase;
return min($permissions);
```
I'll create a PR with that change, if that fixes your issue.

bummzack:
There seem to be other inconsistencies with the `canPurchase` permission checks. The documentation in the `Buyable` interface states that a `ShopBuyableException` should be thrown when the product can't be purchased. That seems to be against the philosophy of permission checks… I guess it was once used to provide some feedback to the user (via Exception message), but I can't find any reference in the code where this is actually being used (`ShopBuyableException` isn't even being thrown anywhere).

I suggest to fix the bug with missing quantity in `1.x` and remove/deprecate `ShopBuyableException` for `2.x`. @markguinn Thoughts?

fonsekaean:
Hello @bummzack, The issue is that the quantity doesnt get passed in to the canPurchase decorator function. I think your code will solve the issue with the canPurchase when the canPurchase decorator has to check for the quantity you are going to purchase. 

Thanks for the fast reply.

markguinn:
Agreed. I think someone went through an exception-happy phase at some point in the past. :) I definitely thing the correct behaviour is to return false rather than throwing an exception. Your suggested approach of fix in 1.0 and deprecate for 2.0 is great.

fonsekaean:
[...]

Issue #90, opened 2012-08-31
Order item validation


jedateach:
It is easy to get items in the cart which shouldn't be in there.

When should items be validated
- Before placing an  order
- Before adding / updating
- On every request?? - can probably avoid this

What should happen to invalid items
- Remove from cart?
- Display a message explaining the problem

Examples
- Required item attributes are not set: size, relation, etc
- Product is no longer available
- Product is now out of stock
- Trying to add a product, but it has variations

bummzack:
There are several flaws with how the Shopping-Cart and Order-Items are updated:

 - No easy way to display custom messages to the user, other than relying on Exceptions and try/catch mechanisms.
 - Preventing/altering item-updates is clunky. There are extension hooks like `beforeAdd`, `afterAdd`, `beforeRemove`, `afterRemove` etc. but these don't provide good tools to intercept unwanted changes (other than throwing an Exception, which doesn't allow to do something like clamping quantity to a max-amount or similar)
 - Too many code-paths that do similar things. An extension that operates on Quantity has to deal with `add` and `setQuantity` (where 2 extension hooks exist per operation). Ideally, item modification would only need one place for validation.
- Switching variations is not affected by any validation. Example: *Variation A* has 5 items, *Variation B* has 10 items in stock. Customer adds 10 items of *Variation B*, can then switch to *Variation A* and there's no way to validate/prevent that change.

bummzack:
We need to think about:

 - A better Cart API with less internal code paths
 - Order-Item validation: Validation of a single order item
 - Cart-validation: Validation of the complete order/cart, including modifiers. There might be scenarios where a modifier relies on specific items to be in cart etc.
 - All changes must be rolled back if validation fails.

The easiest way to implement rollbacks are probably DB transactions. Since we already use DB transactions in `OrderProcessor::placeOrder`, we could use them to roll-back an invalid cart. That would mean that SilverShop will only support Databases with transaction support.

PR #554, opened 2016-10-10
Revert commits that re-added the ShopBuyableException.

bummzack:
Opening this up for discussion for the core-team (@markguinn, @wilr, @anselmdk), or any interested party. 

We have deprecated `ShopBuyableException` and removed it with 2.0. Some of the reasoning can be found in this issue: #443

The gist of it:

- Throwing an Exception from `canPurchase` seemed like a non-intuitive architecture, since the method already tells us whether or not a product can be purchased.
- The only reason the Exception was used there, was to pass a message to the outside.

Since `2.0` was our only chance to clean up API (before 3.0), I'd rather keep the exception removed and look for a short-term and long-term solution to the problem.

Proposed short-term fix:

 - Remove exception throwing from the [silvershop-stock](https://github.com/silvershop/silvershop-stock) module, so that it's compatible with core 2.0
 - Leave it up to the developer to improve the UX where neccessary (eg. limiting the amount of products that can be added to cart) without relying on exceptions

Proposed long-term fix:

There are several solutions for a long-term solution (most of this could be done within the 2.x release I think). I guess we should open a new issue for this and discuss it there. Some ideas:

 - Extend the API of `canPurchase` to pass in some sort of container for messages (reasons why product could not be added). Every extension can then append messages to the container, which can be displayed to the user.
 - Improve UI components, such as the "Add to cart" button or quantity fields so that they can be limited to max-amounts or can be disabled… with the goal to prevent running into invalid situations in the first place

Thanks for reading.

### DIFF
--- a/code/ShopBuyableException.php
+++ b/code/ShopBuyableException.php
@@ -1,9 +1,0 @@
-<?php
-
-/**
- * @package silvershop
- * @subpackage exceptions
- */
-class ShopBuyableException extends Exception {
-    
-}
--- a/code/cart/CartForm.php
+++ b/code/cart/CartForm.php
@@ -43,51 +43,31 @@ class CartForm extends Form
         if (isset($data['Items']) && is_array($data['Items'])) {
             foreach ($data['Items'] as $itemid => $fields) {
                 $item = $items->byID($itemid);
-
                 if (!$item) {
                     continue;
                 }
-
-
-                // delete lines
+                //delete lines
                 if (isset($fields['Remove']) || (isset($fields['Quantity']) && (int)$fields['Quantity'] <= 0)) {
-
-                    try {
-                        ShoppingCart::singleton()->remove($item->Buyable());
-
-                        $removecount++;
-                    } catch(ShopBuyableException $e) {
-                         $form->sessionMessage($e->getMessage(), "bad");
-
-                         break;
-                    }
-
+                    $items->remove($item);
+                    $removecount++;
                     continue;
                 }
-
-                // update quantities
+                //update quantities
                 if (isset($fields['Quantity']) && $quantity = Convert::raw2sql($fields['Quantity'])) {
                     $numericConverter->setValue($quantity);
-
-                    try {
-                        ShoppingCart::singleton()
-                            ->setQuantity($item->Buyable(), $numericConverter->dataValue());
-                    } catch(ShopBuyableException $e) {
-                         $form->sessionMessage($e->getMessage(), "bad");
-
-                         break;
-                    }
+                    $item->Quantity = $numericConverter->dataValue();
                 }
-
-                // update variations
+                //update variations
                 if (isset($fields['ProductVariationID']) && $id = Convert::raw2sql($fields['ProductVariationID'])) {
                     if ($item->ProductVariationID != $id) {
                         $item->ProductVariationID = $id;
-                        $item->write();
                     }
                 }
-
+                //TODO: make updates through ShoppingCart class
+                //TODO: combine with items that now match exactly
+                //TODO: validate changes
                 if ($item->isChanged()) {
+                    $item->write();
                     $updatecount++;
                 }
             }
--- a/code/product/AddProductForm.php
+++ b/code/product/AddProductForm.php
@@ -65,15 +65,9 @@ class AddProductForm extends Form
                 array_intersect_key($data, array_combine($this->saveablefields, $this->saveablefields))
             ) : $data;
             $quantity = isset($data['Quantity']) ? (int)$data['Quantity'] : 1;
-
-            try {
-                $cart->add($buyable, $quantity, $saveabledata);
-
-                if (!ShoppingCart_Controller::config()->direct_to_cart_page) {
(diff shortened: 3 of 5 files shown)


anselmdk:
I'd say let's merge this, and maybe copy your detailed description to an issue for further discussion?

wilr:
Ok, I see but I don't think we should have removed an API in the first place without a medium or long term solution in place and I don't think we should remove it again till a workaround for stock is in place (maybe I'm biased as tons of my shops use stock)

Is there any reason why you think an Exception is not appropriate at this stage? SilverStripe internally uses `ValidationException` for a similar cases.

If the use of an extension is to be avoided canPurchase could return a `ValidationResult` instance and the Cart Form updated to handle all cases (exception, ValidationResult or boolean). 

```
class ProductExtension extends DataExtension {

public function canPurchase($member, $quantity) {
   if($today == 'tuesday') {
      $result = new ValidationResult();
      $result->error('TUESDAY', 'This product is not available to order on tuesdays');

      return $result;
   }
}
```

Cart form would capture any `ValidationResult` instances, group them using `ValidationResult::combineAnd` to present any errors to the user.

If no one else is using stock heavily then happy to submit this as an approach.

bummzack:
Let me clarify

### Architecture 

I think `Exceptions` are not appropriate, since `canPurchase` has a clearly defined meaning. It's like `canView` or `canEdit`, it either returns `true` or `false` which indicates that the product can be purchased or not. The stock module raises an exception instead of returning `true` or `false`, which seems counter-intuitive.

Abusing an `Exception` just to pass a message to the end-user doesn't seem like a good architecture in general.

The way we use the exception is more like as if SilverStripe would throw an Exception from within `canEdit`, just to tell the user *why* he can't edit this DataObject.

### Deprecation and removal

We have deprecated the use of `ShopBuyableException` on March 5th, 2016: https://github.com/silvershop/silvershop-core/commit/de8818b59e444384a3c15bce1b59713d19768e55 (from that point forward, trying to use the exception would have created a deprecation notice)

I think that was plenty of time for users to react to the change until it was removed with the release of `2.0`.

### Short-term solution

What's wrong with removing the Exception from the Stock module? The only drawback of this is, that the user might get a "This product cannot be purchased" message, instead of the "out of stock" message. In the end, the messages both mean pretty much the same, although the latter is of course more specific.

Or, since 2.0.1 was already released (with the exception being back in the codebase), the stock-module could be temporarily constrained to this version.

### Long-term solution

I agree, that we need a better solution, one that also works with other modules that can have an effect on product availability… like I said previously, I'd rather have this discussed in a separate issue, maybe RFC style?

Since `canPurchase` is called via `extend`, returning a complex object such as `ValidationResult` will be troublesome, because we then have to merge these (because we'll get an array of return values, with a value for each extension).

A simple architecture, which would not invalidate our current API (eg, it could go into 2.x) could look like this:

 - Have a container class, eg. `FrontendMessageCollection`. This collection allows adding success and/or error messages. Something like: `$collection->addErrorMessage()`
 - Pass the collection as a third optional argument to `canPurchase`.
 - Each extension can use this parameter to add messages.
 - The invoking class (eg. ShoppingCart or similar) will then look at the return value of `canPurchase` and in case of `false` display the error messages from the collection and in case of `true` display success messages (if any) to the user.

bummzack:
If we're addressing this issue, it might be worthwhile to also look into #90. Maybe `canPurchase` is not the correct place to start making changes… maybe it should be some sort of a cart-items-validation. Something like this could also replace the inflexible messaging code in ShoppingCart (`error`, `message`, `getMessage`, `getMessageType` and `clearMessage` methods).

wilr:
On the short term solution - as long as the user gets prevented from making an order with the stock then I think that would be fine. Nicer messaging can be handled by project developers. 

I was originally alerted to this issue when one of our developers unit tests started failing in that he was able to order products beyond the stock controlled limit (i.e canPurchase did nothing). As long as something stops the execution flow when a product is out of stock then that would be an acceptable base.

Regarding the long term solution I don't think adding more arguments to canPurchase is sustainable and likely going to get us further down a rabbit hole which would be harder to get out of for 3.0. canPurchase being invoked via `extend` is something we define so if complex objects cause issue then we could simply replace the extend() call with a normal function call on all extension instances and handle each result.

bummzack:
Ok, can we agree on the following then?

 1. Merge this PR 
 2. Remove exception throwing from stock-module and ensure tests pass (of course tests need to be modified to just check `canPurchase` and not expect an exception)
 3. Create a new issue or update issue #90 to define an approach how to tackle the issue properly

wilr:
👍  from me

bummzack:
@wilr Ok great. Will you be updating the stock module?

wilr:
I think I'd prefer someone else have a look over it as that would help the module code base and context for #90

bummzack:
I just meant the short term solution, eg. remove throwing of exceptions in the stock module?

wilr:
@bummzack Would still rather someone else review so that it's not just me in the echo chamber trying to talk about stock management. See `StockManagedOrder::beforeSetQuantity` there are a number of places that will need to be tested for stock not just canPurchase.

PR #557, opened 2016-10-11
Wrap all order extension hooks in try/catch blocks.

bummzack:
This allows extensions to interrupt updates to the cart.
Update CartForm to collect/display bad messages from shopping-cart.


bummzack:
The goal of this is to provide a short-term solution to the problem discussed here: #554

PR #561, opened 2016-10-21
Fixed problem with email receipts not being sent after an order has b…

sapiensfio:
Fixed problem with email receipts not being sent after an order has been paid for.


bummzack:
Thanks for reporting and creating a PR!
I've addressed this issue in another PR #563, closing this in favor of the other one.

PR #565, opened 2016-11-29
Correcting the PHPDoc

catharsisjelly:
I assumed the parameter in the function is correct, it was annoying me that the documentation was different from the function


bummzack:
Good catch, thanks!

Issue #569, opened 2016-12-22
Custom Email notifier


catharsisjelly:
I can't see currently a way of adding logic to the `OrderEmailNotifier`. Basically my use case is that the user when purchasing an item (a training course) needs to supply email addresses of people that  are attending the course. The e-mail the attendees get may be different from the person purchasing the training course (with dates etc).

The only way I can see to do this is to build a custom `OrderEmailNotifier` and then use that instead of the current one. This way I can control who get's which template and what information is taken out of the relevant Order.

I'm going to put together a little PR to suggest a way to do this but obviously will take guidance from the community.

bummzack:
It really depends *when* you need to notify attendees. It could be sufficient to just use the `onPaid` extension-hook (implemented in an extension added to `Order`) and send the E-Mails from there.

But I agree, the `OrderEmailNotifier` is not very extensible. An architecture that makes it easier to send all kinds of shop-related Emails could be useful, but it might be hard to find a solution that handles most possible use-cases elegantly.

catharsisjelly:
For my scenario you may want the people attending the course to receive an e-mail before the Order is paid for. For instance some companies ay want to pay by invoice (after X days) so in this case the hook into Paid won't work. I could create a new status of course and hook into that instead.

bummzack:
If you just need to be able to use a custom Subclass of `OrderEmailNotifier`, then this is already in place, see my comment on the PR (#570)

catharsisjelly:
Yep a subclass would do it but was not sure how to make the OrderProcessor use my Subclass and not the standard `OrderEmailNotifier`

bummzack:
You can use the Injector, eg. in your config file:

```yml
Injector:
  OrderEmailNotifier:
    class: MyCustomOrderEmailNotifier
```

Then in your `MyCustomOrderEmailNotifier` class, you could implement other methods, such as `sendNotificationToAttendees` etc.

Call these methods from hooks that you added to an extension to `Order`. You can use the `onPaid` and `onPlaceOrder` hooks, or the `onStatusChange` hook if you need to trigger something on status-change.

catharsisjelly:
Thanks @bummzack I'll cancel my PR and/or change the docs :)

bummzack:
You're welcome. Hope it helps to achieve what you're looking for, otherwise feel free to reopen or create a new PR.

[...]

PR #570, opened 2016-12-22
Added a config hook to use a custom notifier

catharsisjelly:
Looking to supply a solution for #569, also fixed a small bit of documentation

### DIFF
--- a/code/checkout/OrderEmailNotifier.php
+++ b/code/checkout/OrderEmailNotifier.php
@@ -38,7 +38,8 @@ class OrderEmailNotifier
     }
 
     /**
-     * @param $bool
+     * @param bool $bool
+     * @return $this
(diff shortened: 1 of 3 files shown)


bummzack:
This PR is unnecessary, since `OrderEmailNotifier::create`, already calls:

    return Injector::inst()->create('OrderEmailNotifier', $order);

catharsisjelly:
I'm still learning SS so trying to understand the best way to do this, would it be better to use the `->extend`?

bummzack:
Where and how would you use `->extend`?

catharsisjelly:
Obviously not re our conversation :) Hows this for a docs change?

bummzack:
Great addition, thanks.

PR #572, opened 2017-01-04
Setting the Type in constructor

TheBnl:
Setting the Type in the populateDefaults method was not being picked in the value/totals calculation


bummzack:
Thanks!

TheBnl:
Thank you! 👍

PR #574, opened 2017-01-17
Made OrderProcessor extendable

TheBnl:
Made OrderProcessor handle extension hooks, i added `updateGatewayData` because the `transactionId` set here includes special characters (dash). The payment gateway i'm currently using was only accepting alphanumeric characters. Now i can simply add and extension and update the `transactionId` to a format workable for my case.

### DIFF
--- a/code/checkout/OrderProcessor.php
+++ b/code/checkout/OrderProcessor.php
@@ -10,7 +10,7 @@ use SilverStripe\Omnipay\Service\ServiceResponse;
  *
  * @package shop
  */
-class OrderProcessor
+class OrderProcessor extends Object
 {
     /**
      * @var Order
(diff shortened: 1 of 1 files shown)


bummzack:
The reason why the transaction-id gets updated there is, that a failed attempt must have a new transaction-id.

I think you should already be able to achieve what you want by adding an extension to `PurchaseService` and `AuthorizeService` (of the silverstripe-omnipay module) where you update the id via `onBeforePurchase` and `onBeforeAuthorize`… 

May I ask what gateway you're using?

TheBnl:
Thanks for reviewing! Alright, i'll look into that option! I'm using is the [Rabobank](https://github.com/thephpleague/omnipay-rabobank) gateway.

TheBnl:
Adding extensions to the `PurchaseService` and `AuthorizeService` works as you suggested, so i'll close this PR. 

I did need to make an alteration to the data passed to the `onBeforePurchase` method though, it was cast into an array that could not be modified. Casting it into an `ArrayData` object makes it editable. I'll scan trough the PaymentServices and summit a PR there 👍

PR #578, opened 2017-02-10
removed old code from SimpleShippingModifier.

sanderha:
removed old code from SimpleShippingModifier. It now simply looks at orders shipping address country to determine the shipping cost


bummzack:
There seems to be some issue with the Travis setup. I'm going to fix this later.
Thanks @sanderha

PR #580, opened 2017-02-16
added extension hooks to update aditional fields

TheBnl:
Hi, 

I needed to added some extension hooks to update the field data that gets saved in the relation object. Extra fields are needed when for example someone wants to add an inscription to there product.

Firstly the extension hook in `ProductVariation::createItem()` enables me to set extra fields on the relation `$item`.

```php
public function updateCreateItem($item)
{
    $item->Inscription = $this->owner->Inscription;
}
```

Second the hook in `VariationForm::addtocart()` enables me to check if an extra field is set trough the form, and then set it on the `$variation`.

```php
public function beforeVariationAddToCart(VariationForm $form, ProductVariation $variation)
{
    $data = $form->getData();
    if (isset($data['Inscription'])) {
        $variation->Inscription = $data['Inscription'];
    }
}
```

These hooks where needed next to creating the extra field, in this case `Inscription` on the `ProductVariation`, `ProductVariation_OrderItem`.

### DIFF
--- a/code/product/variations/ProductVariation.php
+++ b/code/product/variations/ProductVariation.php
@@ -266,6 +266,9 @@ class ProductVariation extends DataObject implements Buyable
         $item->ProductID = $this->ProductID;
         $item->ProductVariationID = $this->ID;
         //$item->ProductVariationVersion = $this->Version;
+
+        $this->extend('updateCreateItem', $item);
+
         if ($filter) {
             //TODO: make this a bit safer, perhaps intersect with allowed fields
             $item->update($filter);
--- a/code/product/variations/VariationForm.php
+++ b/code/product/variations/VariationForm.php
@@ -121,6 +121,8 @@ class VariationForm extends AddProductForm
                 return json_encode($ret);
             }
 
+            $this->extend('beforeVariationAddToCart', $form, $variation);
+
             if ($cart->add($variation, $quantity)) {
                 $form->sessionMessage(
                     _t('ShoppingCart.ItemAdded', "Item has been added successfully."),


bummzack:
I *could* add your hooks, but I don't believe they're going to improve the situation here.

I've looked at your PR and how the cart handles things and unfortunately discovered several issues…
The issue with your implementation is, that you seem to store the Inscription on the `Product` or the `ProductVariation`, but the `Inscription` should solely be added to the `OrderItem`, since it's specific to each instance of the Product or Variation that's being ordered. With your implementation, you might even run into some sort of race-condition, where a user orders a Product and gets an Inscription from another user.

`AddProductForm` seems to take extra data and save it via here: https://github.com/silvershop/silvershop-core/blob/master/code/product/AddProductForm.php#L64. I was thinking that the cleanest solution would be to add the same logic to the `VariationForm`, but while doing so, I discovered another issue. The additional saveable data is only being applied, when the first product of a type is added to the form. Say you have a customizable Mug. If you order the first Mug, it will save your custom data. If you order the second Mug, you'll just have two Mugs in your cart, but it should be two separate objects!

So it doesn't seem to be as easy as first thought. I'll open a separate issue for this.

TheBnl:
Thanks for looking into this! I'm going to alter my implementation to solely use the `OrderItem` and the `saveablefields` configuration, so to eliminate the possibility of customers ordering each others inscriptions. Then i'll also look into the issue with not being able to order two products of the same type with different inscriptions and continue that discussion on the issue you've opened. #584

bummzack:
@TheBnl I've fixed some issues in shop that prevented this from going to work. See #585 
Here's the setup I've used. Maybe it's useful to you: https://gist.github.com/bummzack/a9bb2dd8949a38cf0299bbbb7b7ad666

bummzack:
I'm closing this PR. If you're running into issues with the new code in master, let me know.

TheBnl:
Awesome thanks for this! 👍

PR #587, opened 2017-02-28
pass the ShopAccountForm $form to the extension to be modified

wildflower:
The Extension call needs a $form or $this passed as a parameter, this is so you can take the $form and change it how you wish.

If you don't pass the $form everything falls over

I'm using my extension method to remove some fields


bummzack:
You can use `$this->owner` in the Extension.

PR #591, opened 2017-04-19
Updated CheckoutStep_Summary.php

AntonyThorpe:
The `updateConfirmationForm`form extension appears twice in the CheckoutStep_Summary class.  Remove from ConfirmationForm function and leave in the summary function, as this this closer to the final return statement.


bummzack:
I think it would be better to remove the *other* `updateConfirmationForm` call. Since `ConfirmationForm` is a public method and should also trigger `updateConfirmationForm` when you call it directly! But you're right, one extend call is enough, but I'd remove the one in the `summary` method.

AntonyThorpe:
Updated as requested.  Thank you.

bummzack:
Thanks Antony!

PR #603, opened 2017-06-09
Placed added to group by

### LINKED ISSUE #602: orderby clause is not in group by sql_mode=only_full_group_by
[User Error] Uncaught SS_DatabaseException: Couldn't run query: SELECT COUNT(*) AS "OrderCount", DATE_FORMAT("Placed",'%d %b %Y') AS "Date" FROM "Order" WHERE "Status" not in ('Cart') AND "Placed" > date_sub(now(), interval 30 day) GROUP BY "Date" ORDER BY "Placed" Expression #1 of ORDER BY clause is not in GROUP BY clause and contains nonaggregated column 'ss_mysite.Order.Placed' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by

tony13tv:
This fixes #602


bummzack:
Thanks @tony13tv!

Issue #605, opened 2017-06-22
CartEditField referencing Product


catharsisjelly:
Hi all,

I'm extending this and have found a possible bug. `CartEditField` in the `editableItems` function is referencing the `Product` class direct. Should this be using a generic function like `Buyable()`? this would occur if I chose to make something which does not extend `Product` use the `Buyable` interface.

For instance if the the think that I am "buying" is a Ticket for an event.

https://github.com/silvershop/silvershop-core/blob/master/code/cart/CartEditField.php#L99

bummzack:
Yes, definitely use `getBuyable()` instead if that doesn't break stuff. 
Maybe the product form could also be improved to be compatible with custom buyables… but a fix is needed there as well: https://github.com/silvershop/silvershop-core/blob/50977db6854f6a2e285cc6c03a73eb81c5f925aa/code/product/AddProductForm.php#L86

catharsisjelly:
This is now merged so am closing

PR #606, opened 2017-06-22
Fix for Issue CartEditField using Product

catharsisjelly:
Fix for #605


catharsisjelly:
I was not sure that this would fix the problem but from my tests it has. It would make me feel safer if another dev can confirm this but for me it fixed what I was working toward.

bummzack:
Buyable should return the Product for the default OrderItems, so this is basically no change for the default shop. And since tests are passing I think this is safe to merge. Thanks @catharsisjelly

PR #608, opened 2017-07-10
Product price should be casted as ShopCurrency

mediabeastnz:
Product price should be casted as ShopCurrency.

It seemed symbol changes only come through in the PriceRange function and not the standard sellingPrice function.

Changing the casting to ShopCurrency seemed to fix this.


mediabeastnz:
Sorry, I've just noticed your Object::useClass... ignore this!

PR #618, opened 2017-10-24
Update composer alias for master branch

dhensby:
I'm pretty sure dev-master shouldn't be an alias for 1.2 and should be for 2.2

Also, you may want to consider making it an alias for 2.x instead of 2.2.x - but I'll leave that up to you.


bummzack:
You're right. Thanks for spotting this.
I think the alias should be `2.x`

dhensby:
updated

PR #644, opened 2018-03-26
fix(ProductVariations): update the SQL table and field names for attr…

gheggie:
…ibute values lookup


bummzack:
Ow, good find. Thanks.

PR #647, opened 2018-04-05
Multi-step checkout documentation fix (to use FQCNs)

gheggie:


bummzack:
This is a documentation issue. Instead of using `CheckoutPage.steps`, one should use the FQCN instead, eg. `SilverShop\Page\CheckoutPage.steps`. I won't change that back, as it would contradict with all other SS4 modules.

Does this PR fix another issue, other than reverting the config name? If yes, please update this PR by using the FQCN. Thanks!

gheggie:
@bummzack Thanks, for that. I'd changed everything to FQCNs except `CheckoutPage` and it wasn't working until I did the previous code changes. Reverting code changes, and making sure everything in config is referenced by FQCN now works. I've updated the PR with just the documentation change.

docs/en/02_Customisation/01_Recipes/Multi_Step_Checkout.md
```diff
@@ -68,7 +68,7 @@ When needing additional Form Fields in a Multi-Step Checkout, your best bet is t
 1) Adjust the yaml config:
 
 ```yaml
-CheckoutPage:
+SilverShop\Checkout\Step\:
```
bummzack: Shouldn't this be `SilverShop\Page\CheckoutPage`?

bummzack:
Thanks a lot!

PR #654, opened 2018-05-01
fix(ShopConfigExtension): update method for parsing AllowedCountries …

gheggie:
…into array (SS4 stores using json_encode)


bummzack:
Wow, I wasn't aware that SS4 stores lists as json-encoded data.
Good catch!

PR #656, opened 2018-05-15
updated visibility of findOrMakeItem

hoodweb:
updated visibility of findOrMakeItem from private to protected, in order to allow for overriding the function in subclasses


bummzack:
Thanks!

PR #665, opened 2018-06-20
Fixes admin interface ss41

gordonbanderson:
With the default config settings I've had to make these changes to get the admin interface not breaking with SS41.

The following were broken:

* Viewing list of orders
* Viewing a single order
* Shop reports where a date range was included.  I've added a config override here, but the default behavior should be the same as it was before.

### DIFF
--- a/example_config.yml
+++ b/example_config.yml
@@ -83,3 +83,7 @@ SilverShop\Cart\ShoppingCartController:
   direct_to_cart_page: true
   # disable security token for cart-links
   disable_security_token: true
+
+SilverShop\Reports\ShopPeriodReport:
+  # IF you are using the silverware/calendar module, uncomment the following line
+  #disable_set_date_format: true
--- a/src/Extension/ShopConfigExtension.php
+++ b/src/Extension/ShopConfigExtension.php
@@ -113,7 +113,7 @@ class ShopConfigExtension extends DataExtension
         asort($countries);
         if ($allowed = $this->owner->AllowedCountries) {
             $allowed = json_decode($allowed);
-            if (!is_null($allowed) && count($allowed) > 0) {
+            if (!empty($allowed)) {
                 $countries = array_intersect_key($countries, array_flip($allowed));
             }
         }
--- a/src/Extension/ShopConfigExtension.php
+++ b/src/Extension/ShopConfigExtension.php
@@ -113,7 +113,7 @@ class ShopConfigExtension extends DataExtension
         asort($countries);
         if ($allowed = $this->owner->AllowedCountries) {
             $allowed = json_decode($allowed);
-            if (count($allowed) > 0) {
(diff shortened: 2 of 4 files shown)


example_config.yml
```diff
@@ -83,3 +83,7 @@ SilverShop\Cart\ShoppingCartController:
   direct_to_cart_page: true
   # disable security token for cart-links
   disable_security_token: true
+
+SilverShop\Reports\ShopPeriodReport:
+  # IF you are using the silverware/calendar module, uncomment the following line
+  #disable_set_date_format: true
```
bummzack: Not necessary to uncomment the line. The comment above is also misleading. The config from `example_config.yml` is **not** being applied!

src/Extension/ShopConfigExtension.php
```diff
@@ -113,7 +113,7 @@ public function getCountriesList($prefixisocode = false)
         asort($countries);
         if ($allowed = $this->owner->AllowedCountries) {
             $allowed = json_decode($allowed);
-            if (count($allowed) > 0) {
+            if (!is_null($allowed) && count($allowed) > 0) {
```
bummzack: I guess we should write:
`if (!is_empty($allowed)) {`
instead. Better readable and more concise.

src/ORM/Filters/MultiFieldPartialMatchFilter.php
```diff
@@ -65,7 +65,7 @@ function ($v) {
             }
         );
 
-        if (count($this->subfilters) > 0) {
+        if (!is_null($this->subfilters) && count($this->subfilters) > 0) {
```
bummzack: Use `!is_empty($this->subfilters)` instead.

gordonbanderson: Will attend to tomorrow

example_config.yml
gordonbanderson: I was changing the behavior, I went with keeping the default the same

bummzack: Yeah, that's fine. But the `example_config.yml` is not being applied, so it doesn't change anything. It could also read:
```yml
# IF you are using the silverware/calendar module, apply the following to your config
SilverShop\Reports\ShopPeriodReport:
  disable_set_date_format: true
```

gordonbanderson: I'm not clear how you wish me to proceed here.  Ping me on silverstripe-users slack, my handle is nontgor

bummzack: No problem. We can leave it the way you suggested. I'm just saying that it doesn't really matter, since that file doesn't have any effect, so we could add a complete example and add a comment that the user should copy this into his config to make it silverware/calendar compatible.

bummzack:
Thank you!

PR #669, opened 2018-07-02
Redirect to the next step if validation is passed

anhld:
With `skip_if_logged_in` = `true` , it should be redirected to the next step if validation is passed.


bummzack:
Thanks!

PR #672, opened 2018-08-05
Loosen CMS constraint (Fixes #671)

### LINKED ISSUE #671: Module should require specific core components, not recipe-cms

wilr:


wilr:
Ok so tests fail as they assume a `Page` class. Can make the recipe `require-dev`.

kinglozzer:
Thanks @wilr! My personal preference would be to not require the recipe at all and instead require the installer in `.travis.yml` in the same way that the [blog module does](https://github.com/silverstripe/silverstripe-blog/blob/master/.travis.yml#L28), but I’ll leave that call to you guys 😄

kinglozzer:
@wilr I guess the current Travis failures are because `phpunit.xml.dist` needs to be updated to point to the CMS bootstrap file: https://github.com/silverstripe/silverstripe-blog/blob/master/phpunit.xml.dist#L1

wilr:
Failing tests now look pre-existing so going to merge

PR #678, opened 2018-08-28
Remove use of SHOP_DIR

wilr:


liamchowneydev:
Whats the latest with this bug? Can I use composer to fix my instance?

wilr:
Yes it's been merged and is available in the `2.x-dev` release. It will make a stable release number in the future.

liamchowneydev:
I have installed the update version and the shop dir element is fixed and acts the same as the temporary fix I had in place. I still get this error when you click on the already existing address checkbox
[Emergency] Uncaught TypeError: Argument 1 passed to SilverShop\Shipping\Model\Zone::cache_zone_ids() must be an instance of SilverShop\Model\Address, null given, called in /home/oneweigh/public_html/vendor/silvershop/shipping/src/Extension/OrderShippingExtension.php on line 98

PR #680, opened 2018-09-11
Test SilverShop 2.x / SS3.7 with PHP7.1

wernerkrauss:


bummzack:
Good idea. What about 7.2? Isn't that more important than the change from 7.0 to 7.1?

wernerkrauss:
I tried 7.2 locally but it broke due to PHPUnit3. Let's test if it works on travis ;)

bummzack:
Ah yes. I remember that I tried that at some point as well… didn't work on travis either :(

wernerkrauss:
Ping, I removed my latest commit, so everything is green now.

bummzack:
Thank you!

PR #685, opened 2018-10-13
publish the image when saved

sparkcom:


wilr:
@sparkcom you shouldn't have to publish the file manually - it's declared as an `$owns` object. Just check that your publishing code does `$product->publishRecursive()` rather than a plain `publishSingle()`

PR #698, opened 2018-12-05
fix resolving of index template for stepped checkout

xini:

PR #699, opened 2018-12-05
CheckoutPage_membership should be treated as Layout

wernerkrauss:
Without explicitely using "Layout" in the template name it gets wrapped
in global Page.ss which is the desired behaviour


wernerkrauss:
Seems #698 is the cleaner solution to this problem

PR #706, opened 2019-01-08
casting of ShopCurrency

wernerkrauss:


wilr:
Test failures are pre-existing and unrelated. Will see if I can get ontop of those failures today.

PR #718, opened 2019-06-12
Hook to update uniquedata

CMeldgaard:
Added extend point to make it possible to perform additional unique data checks on the orderitem.


wilr:
Thanks!

PR #719, opened 2019-08-06
Include result in params to extension method in ShoppingCartControlle…

gheggie:
…r::add()


wilr:
Merge failure is pre-existing. I'll resolve that.

PR #722, opened 2019-09-18
Exclude deprecations from custom error handler

Greg808:
Fixes an error, where no emails / receipts are sent, when an order is paid (via payment provider). We have local mail set up , no SMTP. The problem is, that Swift standard transport is deprecated and throws a E_USER_DEPRECATED error which is now ignored.


wilr:
Thanks! Will review the backlog and tag a release.

PR #724, opened 2019-10-31
FIX: fix bug preventing the view of each order record in CMS

mikenuguid:
- update $searchable_fields
- include get prefix to methods used in configs


wilr:
Thanks

mikenuguid:
Thanks @wilr!

PR #725, opened 2019-11-01
Use Label as title for frontend dropdown field

RVXD:
In the frontend dropdown the Label field should be used instead of the Name field.
This is mentioned in the source:

```
    private static $db = [
        'Name' => 'Varchar', //for back-end use
        'Label' => 'Varchar' //for front-end use
    ];
```


wilr:
Thanks!

PR #726, opened 2019-11-07
Use IntlLocales to get the localized country name

TheBnl:
Use IntlLocales to get the localized country name. This saves having to keep a file with translated country names.

### DIFF
--- a/src/ORM/FieldType/ShopCountry.php
+++ b/src/ORM/FieldType/ShopCountry.php
@@ -23,11 +23,7 @@ class ShopCountry extends DBVarchar
      */
     public function Nice()
     {
-        $val = ShopConfigExtension::countryCode2name($this->value);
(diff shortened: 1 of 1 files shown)


src/ORM/FieldType/ShopCountry.php
```diff
@@ -23,11 +23,7 @@ public function forTemplate()
      */
     public function Nice()
     {
-        $val = ShopConfigExtension::countryCode2name($this->value);
-        if (!$val) {
-            $val = $this->value;
-        }
-        return _t(__CLASS__ . '.' . $this->value, $val);
+        return IntlLocales::singleton()->countryName($this->value);
```
wilr: Does this need to be imported?

TheBnl: It definetly does!

PR #728, opened 2019-11-18
Order Link function should point to account if logged in

RVXD:
Changed the order so that if Logged in user is sent to account page instead of checkout page.


wilr:
Thanks!

PR #729, opened 2019-11-22
Add unique modifiers from the config

shoosah:
Currently if modifier is already defined in a third party config, defining the modifier in the main app cause errors. This pull request will make sure that only unique set of modifiers from the config is used.


wilr:
Thanks!

PR #735, opened 2020-01-21
Misc fixes made during project

madmatt:
During a recent upgrade project, we reviewed an old fork that we were on of silverstripe-shop (the precursor to silvershop).

During this, I found the following changes that were missing from silvershop-core, there are four distinct things that we've added / resolved (each commit is a single fix/change that we've made).

The four changes are:
- Fix OrderProcessor not setting the paid date when an order has nothing outstanding
- Show any OrderModifiers added to the order when viewing order details in CMS
- Allow order total calculation to always include any applied discounts/coupons
- Provide a Link method for product variations that links to the product page

I'm happy to raise these as individual pull requests if you'd prefer, but thought that all four are relatively benign so it made sense to just raise them at the same time. Let me know if you'd prefer I separate them.

Thanks!

### DIFF
--- a/src/Checkout/OrderProcessor.php
+++ b/src/Checkout/OrderProcessor.php
@@ -291,7 +291,6 @@ class OrderProcessor
             $this->order->Status = 'Unpaid';
         } else {
             $this->order->Status = 'Paid';
-            $this->order->Paid = DBDatetime::now()->Rfc2822();
         }
 
         if (!$this->order->Placed) {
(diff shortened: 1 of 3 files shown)


wilr:
Thanks @madmatt yeah one is fine from the looks, all small. The test failure looks related to your changes though so if you could review and update test if required.

madmatt:
Thanks @wilr, I've fixed the test failure now by reverting one of the changes (found out that the `Paid` datetime is actually set, but it's done elsewhere now, and the receipt email isn't sent if the paid date already exists 🤦‍♂ 

Great test suite to pick that up!

madmatt:
Hey @wilr, any chance you'd be comfortable merging this now that the tests are fixed?

I've got a couple more things coming shortly hopefully as well.

wilr:
Tests are green so can merge :) thanks for updating! Once the other fixes come in I'll tag a new release.

madmatt:
Awesome, thanks Will!

PR #737, opened 2020-02-13
FEAT: Allow OrderStatusLog objects to *not* be forcibly linked to an order

madmatt:
As per description. Added the following the the changelog which describes the feature:

Add new `order_is_required` config flag to `OrderStatusLog`, set to true by default (for backwards-compatibility). Set it to `false` to allow `OrderStatusLog` objects to be created that are not linked to any particular `Order` object.

This change is backwards-compatible to 3.0.0.


madmatt:
Thanks Will!

PR #739, opened 2020-03-02
FIX: Replace Image::Fit() with Image::Pad()

mikenuguid:
- per Silverstripe 4.0 Filesystem changelog: https://docs.silverstripe.org/en/4/changelogs/4.0.0/#overview-filesystem, Image::SetSize() should be replaced with Image::Pad()
- fixes incorrect image size displayed on shop


mikenuguid:
Thanks @wilr

PR #740, opened 2020-03-06
Store version of ProductVariation in order

RVXD:
This function was missing. The version of the ProductVariation should be stored on placement of order.


wilr:
Thanks!

PR #744, opened 2020-04-21
Toast/custom admin notification

intotheweb101:
adding in the option to have a CMS controlled email for admin notifications for admin emails

### DIFF
--- a/code/checkout/OrderEmailNotifier.php
+++ b/code/checkout/OrderEmailNotifier.php
@@ -90,7 +90,12 @@ class OrderEmailNotifier
         $email = $this->buildEmail($template, $subject);
 
         if ($copyToAdmin) {
-            $email->setBcc(Email::config()->admin_email);
+            if($adminEmail = SiteConfig::current_site_config()->AdminEmail){
+                $email->setBcc($adminEmail);
+            }else{
+                $email->setBcc(Email::config()->admin_email);
+            }
+            
         }
         if ($this->debugMode) {
             return $email->debug();
@@ -128,9 +133,15 @@ class OrderEmailNotifier
             '',
             array('OrderNo' => $this->order->Reference)
         );
-
-        $email = $this->buildEmail('Order_AdminNotificationEmail', $subject)
-            ->setTo(Email::config()->admin_email);
+    
+        if($adminEmail = SiteConfig::current_site_config()->AdminEmail){
+            $email = $this->buildEmail('Order_AdminNotificationEmail', $subject)
+                ->setTo($adminEmail);
+        }else{
+            $email = $this->buildEmail('Order_AdminNotificationEmail', $subject)
+                ->setTo(Email::config()->admin_email);
+        }
+       
 
         if ($this->debugMode) {
             return $email->debug();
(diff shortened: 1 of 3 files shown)


code/cms/ShopConfig.php
```diff
@@ -42,7 +43,9 @@ public function updateCMSFields(FieldList $fields)
                         _t("ShopConfig.CustomerGroup", "Group to add new customers to"),
                         "Group"
                     ),
-                    UploadField::create('DefaultProductImage', _t('ShopConfig.DefaultImage', 'Default Product Image'))
+                    UploadField::create('DefaultProductImage', _t('ShopConfig.DefaultImage', 'Default Product Image')),
+                    TextField::create('AdminEmail', 'AdminEmail')
```
wilr, line 47: Can you wrap translate function around this AdminEmail label (see above).

code/cms/ShopConfig.php
```diff
@@ -8,6 +8,7 @@ class ShopConfig extends DataExtension
 {
     private static $db      = array(
         'AllowedCountries' => 'Text',
+        'AdminEmail' => 'Text',
```
wilr, line 11: This doesn't need to be a Text field, `Varchar(150)` should suffice.

code/checkout/OrderEmailNotifier.php
```diff
@@ -164,18 +175,33 @@ public function sendReceipt()
...
+                _t(
+                    'ShopEmail.CancelSubject',
+                    'Order #{OrderNo} cancelled by member',
+                    '',
+                    array('OrderNo' => $this->order->Reference)
+                ),
+                $this->order->renderWith('Order')
+            );
+        }
         $email->send();
```
wilr, line 205: This is a lot of duplicate code which does the same thing. You're only changing the admin email so set that first too a variable and remove the unneeded duplication.

code/checkout/OrderEmailNotifier.php
```diff
@@ -202,7 +228,12 @@ public function sendStatusChange($title, $note = null)
...
             $adminEmail = Config::inst()->get('OrderProcessor', 'receipt_email');
         } else {
-            $adminEmail = Email::config()->admin_email;
+            if(SiteConfig::current_site_config()->AdminEmail){
+                $adminEmail = SiteConfig::current_site_config()->AdminEmail;
+            }else{
+                $adminEmail = Email::config()->admin_email;
+            }
+            
         }
```
wilr, line 237: Since you have this logic in 2x places, I would recommend creating a function which handles this

```
public function getAdminEmail() {
  $adminEmail = SiteConfig::current_site_config()->AdminEmail;
  
  if (!$adminEmail) {
    $adminEmail = Email::config()->admin_email;
  }

  return $adminEmail;
}
```

Then use this function instead of the if / else check used here and above.

intotheweb101:
@wilr I have pushed some updates to this now, Thanks for the feedback: once you are happy I will make a pull request to allow the same for the ss4 version

code/cms/ShopConfig.php
intotheweb101: Wrapped in Translate

wilr:
Thanks!

PR #754, opened 2020-12-07
Updated Multi_Step_Checkout.md

AntonyThorpe:
Added correct path to the `CheckoutStepShippingMethod` class.  Thanks.


bummzack:
Thanks, good catch!

PR #755, opened 2020-12-09
Updated SilverShop\Checkout\Component\Member.php

AntonyThorpe:
With the stepped checkout process, a new member can be created at the beginning of the checkout process.  If the customer does not complete the checkout process, we end up with a member in SilverStripe without a group.  Therefore, add the Customer Group to a new member immediately after creation.  Thank you.


wilr:
Thanks!

PR #757, opened 2020-12-27
Get the dbObject instance of the PurchaseComplete message

TheBnl:
By passing the HTMLText object the text renders properly in the email


bummzack:
Good find, thanks!

PR #769, opened 2021-05-20
Allow overriding of latest email for order

oliver-norden:
This will allow for a nieche scenario where a user may desire to change their email used for order notifications without affecting the email used for logging in, payments etc.


wilr:
Thanks!

PR #771, opened 2021-09-22
Attribute dropdownField should be using Label instead of Name as its …

RVXD:
I guess branch '2' is de base branch for SilverStripe 3 sites.


wilr:
Yes, ideally you'd upgrade to SS4 soon! Silverstripe 3 is quite out of date now.

PR #773, opened 2021-10-14
Cleanup: Remove useless Product::Link() method

### LINKED ISSUE #772: Product::Link() is useless
In SS4.x, `Product::Link()` is useless, as it only calls `parent::Link()` and `$this->extend(updateLinks)`. The latter is already called in `parent::Link`, and when an `updateLink()` method in an extension doesn't check, if the link is already updated, it updates the Link again - which might result in invalid URLs.

wernerkrauss:
as it doesn't provide any custom value over `parent::Link()`

fixes #772

### DIFF
--- a/src/Page/Product.php
+++ b/src/Page/Product.php
@@ -442,13 +442,6 @@ class Product extends Page implements Buyable
         return false;
     }
 
-    public function Link($action = null)
-    {
-        $link = parent::Link($action);
-        $this->extend('updateLink', $link);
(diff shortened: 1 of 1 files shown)


Marimchen:
Can you please tag a patch for the fix, or tag it in a release?   
Thank you very much

wilr:
https://github.com/silvershop/silvershop-core/releases/tag/3.0.4

PR #777, opened 2022-03-11
fixed the sales report so it returns the number of sales and revenue …

fonsekaean:
…generated during a certain period of time.

This report was returning the products created within a certain window of time than the product's performance as per its description.


wilr:
thanks!

PR #790, opened 2023-09-27
Added nojquery js for addressbook display logic

BettinaMaria98:
- Added js file with vanilla javascript to avoid using jquery. 
- Updated AddressBook Component to load CheckoutPage.nojquery.js file instead jquery if no jquery file is defined.


wilr:
Thanks! Great, nowadays we could probably drop the jQuery one, not sure what benefit that provides over plain js

PR #792, opened 2023-10-16
 AdressBook Update: add address type to newaddress-option

MLindenhofer:
Hi, 
I made an update to the following line in order to make the address type show up in the "Create new address" - option during the Checkout Process. 

https://github.com/silvershop/silvershop-core/blob/6e0c303ce0f6fe989282d35350adf90fcf260cc7/src/Checkout/Component/AddressBook.php#L75

Best regards


wernerkrauss:
@wilr do we still use transifex for creating / managing translations? If yes, then we need a new translation key for the changed translation.

Apart from that it looks good to me. Thanks for the PR, @MLindenhofer

PR #796, opened 2023-10-19
Update AddressBook.php SS4

MLindenhofer:
Update AddressBook translation string by adding the addresstype -  for SS4 version.


wernerkrauss:
Thanks for the PR for SS4 branch

PR #800, opened 2023-10-27
CI: omnipay doesn't have a master branch any more

wernerkrauss:
See #795


wernerkrauss (commented):
Try other version for silverstripe-omnipay

PR #810, opened 2023-12-13
Update ProductOrderItemTest.php

AntonyThorpe:
set $beachBall variable to lowercase. Thanks.

### DIFF
--- a/tests/php/Model/Product/ProductOrderItemTest.php
+++ b/tests/php/Model/Product/ProductOrderItemTest.php
@@ -32,7 +32,7 @@ class ProductOrderItemTest extends FunctionalTest
     /**
      * @var Product
      */
-    protected $beachBall;
+    protected $beachball;
 
     /**
      * @var Product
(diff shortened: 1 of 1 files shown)


wernerkrauss:
Hi @AntonyThorpe thanks for that PR ;)

Do you think, `mp3player` should also be convertet into `mp3Player`?

AntonyThorpe:
Hi @wernerkrauss, depends on the coding convention followed by this Repository. PSR-1 recommends a consistent way of declaring properties so one way needs to be chosen.  Using camelCase would be consistent when shifting back and forth between PHP & javascript so that would help.  To answer your questions, yes.  Thanks.

wilr:
Thanks @AntonyThorpe sorry yes I was trying to clean these up over the weekend to fix the lint errors in VSCode. 

Looks like we have tests working again! Yay. Looking at the logs, I think fixing Onimay is next on my list, there is a few warnings and errors in PHP 8+

Issue #779, opened 2022-08-26
Product's BasePrice form field in backend not configurable for e.g. German decimals


wernerkrauss:
In many parts of Europe (e.g. Germany, Denmark, Austria) we use the comma as a decimal seperator; prices are written like 7.000,00 and not like in English speaking countries e.g. 7,000.00

This should be configurable somehow, e.g. by using NumericField instead of Textfield for BasePrice.

PR #814, opened 2024-01-02
Fix currency formatting

wernerkrauss:
somehow related to #779 some more formatting for different prices in the backend.

For formatting payments see https://github.com/silverstripe/silverstripe-omnipay/issues/244

@wilr and/or @forsdahl please have a look at this PR and merge if possible. I disabled OrderModifier's Amount() method which should be fine IMHO but could possibly break some stuff.

### DIFF
--- a/client/dist/css/shopcms.css
+++ b/client/dist/css/shopcms.css
@@ -43,3 +43,7 @@ Integrated webfont. Created via http://fontello.com/
   .shop-order__address {
     vertical-align: top;
   }
+
+  td.col-TotalNice {
+    text-align: right;
+  }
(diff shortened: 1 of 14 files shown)


wilr:
All checks are green so I'm happy

chromos33:
disabling the Amount method breaks the ShowInTable() function in OrderDiscountModifier in the Discount Module.

wernerkrauss:
@chromos33 Thanks for reporting this. Can you please try changing OrderDiscountModifier's ShowInTable() method to:

```
    public function ShowInTable()
    {
        return $this->Amount > 0;
    }
```

so using the magic DB-Field value directly instead of the getter method

PR #817, opened 2024-02-09
Don't allow archiving of a category if it has products assigned to it.

### LINKED ISSUE #803: canArchive on ProductCategory
If you archive a ProductCategory, all products that are children of this ProductCategory, will also be archived.
This is Silverstripe standard behaviour for pages, but it might not be transparent to the user, when working with products in multiple categories. 

Example: The user has "Blue t-shirt" product in the category: "New T-shirts" but also the product has an additional category "Coloured T-shirts". 

The user now deletes the category "New t-shirts", expecting the product to still exist in "Coloured T-shirts", but it does not, it is now archived.

wernerkrauss:
Fixes #803

### DIFF
--- a/src/Page/ProductCategory.php
+++ b/src/Page/ProductCategory.php
@@ -6,6 +6,7 @@ use Page;
 use SilverShop\Extension\ProductVariationsExtension;
 use SilverStripe\i18n\i18nEntityProvider;
 use SilverStripe\ORM\DataList;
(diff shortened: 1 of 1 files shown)


wilr:
@wernerkrauss Looks good. Probably a candidate for a unit test case but all good to merge to me. After merging this did we want to merge it up to `main`?

PR #828, opened 2024-06-13
Shift private static variables from yaml file to classes

AntonyThorpe:
Some variables are declared in config.yml.  Shift them into the classes so that they can be found :).  Thanks.


wilr (approved):

AntonyThorpe:
Thanks for approving my PR @wilr.  What happens next please?

PR #847, opened 2025-09-29
fix response when increasing number of item in cart is successful

xini:
The response object should be a `HTTPResponse`, and only be set in the case of an error. 

If increasing the number of items is successful, the response should be `null`. 

If the response is given (!= `null`), the `redirect` method will try to redirect to that response. In this case, because the response is set to a string, we get a white screen of death.

This issue was introduced by https://github.com/silvershop/silvershop-core/commit/3427d6327b3d158322a6cf32df395602e09031f1#diff-adc2d15f712ffc2573c65cfa5e8217efeaf1bd035ed5efed08be0658d0e4d878R210.

### DIFF
--- a/src/Cart/ShoppingCartController.php
+++ b/src/Cart/ShoppingCartController.php
@@ -197,9 +197,7 @@ class ShoppingCartController extends Controller
 
             $result = $this->cart->add($product, $quantity, $request->getVars());
 
-            if ($result) {
-                $response = $this->cart->getMessage();
-            } else {
+            if (!$result) {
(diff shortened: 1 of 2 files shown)


wilr:
Thanks @xini will merge once tests are passing. May require updates ?

xini:
Thanks @wilr. I don't like changing tests to suit a PR ;) Can you confirm my reasoning that the `$response` variable should be null unless there is an error? If so, I'll adjust the tests to match that behaviour. I think that commit I referenced changed both the tests as well as the `$response'.

PR #853, opened 2025-11-25
OrderItem.php: update Image method to allow the return of null

AntonyThorpe:
Change the return type of the Image method to allow null.  Thanks.


wilr:
Thanks!

PR #870, opened 2026-05-30
Replace boolean with bool

AntonyThorpe:
In php 8.5, non-canonical cast (boolean) is deprecated, use the (bool) cast instead.  Thanks.


wilr:
Thanks!

PR #889, opened 2026-09-18
Fix infinite recursion / OOM in ShoppingCart::current() when a stock check re-enters the cart

RVXD:
### Problem
`current()` runs `removeUnavailableItems()` on every call, which calls each item's `canPurchase()`. A `Buyable` stock extension's `canPurchase()` commonly inspects the cart via `ShoppingCart::curr()` → `current()` → `removeUnavailableItems()` → `canPurchase()` → … → **infinite recursion**, exhausting memory on the cart page. This surfaced after #875 added the prune-on-load into `current()`.

### Fix
Add a re-entrancy guard (`$removingUnavailable`) around the prune, so a nested `current()` returns the already-loaded order instead of re-pruning. The prune still runs once per top-level `current()` call — no behaviour change otherwise, and no API change.

### Notes
Affects any SS6 site whose products read the cart during `canPurchase()` (e.g. `silvershop/stock`). Refs #875 (introduced the prune-on-load), #888 (prior SS6 cart fix).


wilr:
Good catch!

